PR #1, opened 2024-07-26
open-api-spec for F2

bjoydeep:
This contains the OpenAPI spec for F2 (cluster and policy as well as relationships). This also contains the Host - data, but not changing this thru this PR. 

Refer: https://issues.redhat.com/browse/RHCLOUD-34105

@csams I believe you are working to add the proto spec for Hosts?


bjoydeep:
we may want to have a generic file name. Version should not be buried in the spec name?

PR #8, opened 2024-08-05
Change "tags" to "labels" per UX

randymgeorge:


csams (changes requested):
The pb level changes look good!

Run
```bash
# install dependencies for working on the api
make init

# generate the local go grpc and http clients
make api

# to find any errors in the service logic the change may have introduced.
make build
```

There are a couple of places in https://github.com/project-kessel/inventory-api/blob/main/internal/service/common/resource.go that will need to update to absorb the changes to keep the service running.  The model in biz/common also would need to update if we want to propagate that far at this point.

PR #9, opened 2024-08-05
Update health conventions, regenerate

alechenninger:


csams (changes requested):
Other code in the project already uses some of these.  Run `make build` to find the errors.

alechenninger:
I fixed a bunch of errors in #10 that I forgot to fix before merging those updates. I'll fix these here and #10  should cover the rest.

PR #13, opened 2024-08-06
fix go lint issues and add github workflow

akoserwal:
- Fix golang lint issues
- Add github workflow

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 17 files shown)


csams (commented):
The lint changes are great and much appreciated.

* Do any of the workflows prevent merges?  (CodeQL, buf breaking changes, etc.)
* Is CodeQL a "blessed" service?
* Do we need to align with any workflow automation used by OCM / OpenShift?

akoserwal:
I reused what we configured for the relations API to be consistent: https://github.com/project-kessel/relations-api/tree/main/.github/workflows.

I haven't seen any major challenges with code merges due to CodeQL. if you have any suggestions, we can align with ocm/openshift workflows

PR #16, opened 2024-08-06
Draft more idiomatic proto for relationship reference

alechenninger:


alechenninger (commented):
These needs some thought as to how this should be mapped to HTTP

jmelis:
@alechenninger I'm going to close this one just to keep things clean. Please ping me if you still think we should do this, I'm lacking context for this, and it's outdated.

PR #30, opened 2024-08-09
Pin dependencies in ./api/Dockerfile

csams:
Pin versions to give consistent API generation and project builds instead of using `@latest`, which doesn't.


josejulio (approved):
This looks good! This will ensure consistent builds.

I think we can pin deps using `go.mod` file and that might get us automation such as `dependabot` to bump them, but unsure if that's something we want to pursue right now.

PR #42, opened 2024-08-16
Adds build flag to compose calls to prevent stale executions 

josejulio:
## Describe your changes

Every time we call `make inventory-up` (or the sso version), it will check if there has been any change in source code to decide if we need to re-build because of those changes, this should help to run any stale build.

- Adds  `build` flag to compose excecutions - this prevents from using old builds if there are changes in the code.
- Also made more granular the required files in order to build - this will prevent from non-required files to trigger a full dependency download / build - slowing us down.

Fixes #

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,19 @@ RUN ln -s /usr/local/go/bin/go /usr/local/bin/go
 
 WORKDIR /workspace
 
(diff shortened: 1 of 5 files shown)


csams (approved):
lgtm

PR #46, opened 2024-08-19
[RHCLOUD-34462] Rename data fields to resource_data in proto files

csams:
Rename `data` fields to `resource_data` in proto files for consistency.


josejulio (approved):
lgtm

PR #48, opened 2024-08-19
RHCLOUD-34569 - Ensure call to register are valid requests

josejulio:
## Describe your changes

- ...

Fixes #

## Checklist


api/kessel/inventory/v1beta1/rhel_host.proto
```diff
@@ -14,14 +14,15 @@ option java_package = "org.project_kessel.api.inventory.v1beta1";
...
+      [ (google.api.field_behavior) = INPUT_ONLY, (validate.rules).message.required = true ];
 
-  // The entities that registered this item in the Kessel Asset Inventory. The
-  // same resource may be registered by multiple reporters.
-  repeated ReporterData reporters = 353323086
-      [ (google.api.field_behavior) = OUTPUT_ONLY ];
+  // Todo: This is no longer required?
+//  // The entities that registered this item in the Kessel Asset Inventory. The
+//  // same resource may be registered by multiple reporters.
+//  repeated ReporterData reporters = 353323086
```
josejulio, line 26: @randymgeorge I saw reporters field being removed from other resource in your previous PR. 

Are we removing it from all resources?

randymgeorge: This array was removed for on-premise which is using the REST apis.  I do not see the need for this until we have a query in place.  It is not a response to a CRUD operation.

api/kessel/inventory/v1beta1/policy_detail.proto
```diff
@@ -21,5 +23,5 @@ message PolicyDetail {
   // Defines if the policy is currently enabled or disabled across all targets
   bool disabled = 270940796;
```
josejulio, line 24: Note that this is still not correctly modelled. 
Booleans will default to false, but is not possible (as is) to test if they are set or not.

I saw that we might want to default, so it might be fine if that's the case.

internal/service/hosts/hosts.go
```diff
@@ -29,12 +26,7 @@ func New(c *biz.HostUsecase) *HostsService {
 
 func (c *HostsService) CreateRhelHost(ctx context.Context, r *pb.CreateRhelHostRequest) (*pb.CreateRhelHostResponse, error) {
 	if err := r.ValidateAll(); err != nil {
-		return nil, err
-	}
-
-	//TODO: refactor / abstract resource type strings
-	if !strings.EqualFold(r.Host.Metadata.ResourceType, biz.ResourceType) {
```
josejulio, line 36: Removing this validation, `ResourceType` is an output only data so it doesn't make sense to test as input.

internal/service/hosts/hosts.go
```diff
@@ -69,20 +61,16 @@ func (c *HostsService) DeleteRhelHost(ctx context.Context, r *pb.DeleteRhelHostR
 }
 
 func hostFromCreateRequest(r *pb.CreateRhelHostRequest, identity *authnapi.Identity) (*biz.Host, error) {
-	if identity.Principal != r.Host.ReporterData.ReporterInstanceId {
-		return nil, errors.Forbidden("FORBIDDEN", "Reporter identity must match the provided reporter instance identity")
```
josejulio, line 73: Same as above, `ReporterInstanceId` is readonly (output only)

internal/service/notificationsintegrations/notificationsintegrations.go
```diff
@@ -29,12 +26,7 @@ func New(c *biz.NotificationsIntegrationUsecase) *NotificationsIntegrationsServi
 
 func (c *NotificationsIntegrationsService) CreateNotificationsIntegration(ctx context.Context, r *pb.CreateNotificationsIntegrationRequest) (*pb.CreateNotificationsIntegrationResponse, error) {
 	if err := r.ValidateAll(); err != nil {
-		return nil, err
-	}
-
-	//TODO: refactor / abstract resource type strings
-	if !strings.EqualFold(r.Integration.Metadata.ResourceType, biz.ResourceType) {
-		return nil, errors.BadRequest("BADREQUEST", fmt.Sprintf("incorrect resource type: expected %s", biz.ResourceType))
+		return nil, errors.BadRequest("BADREQUEST", err.Error())
```
josejulio, line 29: Same as above

internal/service/notificationsintegrations/notificationsintegrations.go
```diff
@@ -69,20 +61,20 @@ func (c *NotificationsIntegrationsService) DeleteNotificationsIntegration(ctx co
 }
 
 func notificationsIntegrationFromCreateRequest(r *pb.CreateNotificationsIntegrationRequest, identity *authnapi.Identity) (*biz.NotificationsIntegration, error) {
-	if identity.Principal != r.Integration.ReporterData.ReporterInstanceId {
-		return nil, errors.Forbidden("FORBIDDEN", "Reporter identity must match the provided reporter instance identity")
```
josejulio, line 73: Same as above

api/kessel/inventory/v1beta1/policy_detail.proto
randymgeorge: This should be fine. Disabled = false would be a good default.

randymgeorge (approved):

PR #49, opened 2024-08-21
RHCLOUD-34760 | refactor: update resource name for "notifications/integration"

MikelAlejoBR:
## Describe your changes
The SpiceDB schema for Notifications names the integrations as a combination of the "notifications" namespace and the "integration" resource name, resulting in the following full resource name: "notifications/integration". The goal is to represent that in the inventory API.

Fixes [[RHCLOUD-34760]](https://issues.redhat.com/browse/RHCLOUD-34760)

## Checklist

### DIFF
--- a/internal/biz/notificationsintegrations/notificationsintegrations.go
+++ b/internal/biz/notificationsintegrations/notificationsintegrations.go
@@ -7,7 +7,7 @@ import (
 )
 
 const (
-	ResourceType = "notifications-integration"
+	ResourceType = "notifications/integration"
 )
 
 // NotificationsIntegration is a NotificationsIntegration repo.


randymgeorge:
The names have all been snake_case.  If you used the /, then the URL would be api/inventory/v1beta1/notifications/integration and the resource type would be integration.

MikelAlejoBR:
Thakn you for the review, @randymgeorge ! This came from me using the client to try to create an integration, and the current API required me to specify a resource type. Since it didn't match, I assumed that we needed to update it. However, per [RHCLOUD-34642](https://issues.redhat.com/browse/RHCLOUD34642), we no longer will require to specify the resource type. Also, per your comments I guess that the resource type we specify in the request and the one that is present in the SpiceDB schema is different, so I think we can close this PR.

PR #60, opened 2024-08-26
Replace http client with gRPC for kessel relations API and add token …

akoserwal:
…generation option with gRPC

### PR Template:

replace http client with gRPC for Kessel relations API integration and add token refresh logic with enable/disable token authentication

## Describe your changes

- ...

Fixes #

## Checklist


alechenninger:
@akoserwal @jmelis Previous feedback from Joydeep was that this needed to be HTTP1.1 for on prem compatibility (e.g. if there is some reverse proxy that does not support HTTP2 in the way) – are we okay w/ HTTP2?

I would much prefer gRPC (for internal communication especially), and I somewhat doubt it would be a problem, but just want to make sure it's not going to surprise anyone.

jmelis:
> @akoserwal @jmelis Previous feedback from Joydeep was that this needed to be HTTP1.1 for on prem compatibility (e.g. if there is some reverse proxy that does not support HTTP2 in the way) – are we okay w/ HTTP2?
> 
> I would much prefer gRPC (for internal communication especially), and I somewhat doubt it would be a problem, but just want to make sure it's not going to surprise anyone.
> 
> Slack thread for posterity: https://redhat-internal.slack.com/archives/C06DSLVDHM2/p1724779130814169

Good point @alechenninger. We will add a knob eventually for on-prem, but I think we should make gRPC the default transport for hosted.

jmelis (changes requested):
some minor changes

akoserwal:
@jmelis I verified the PR by running a locally keycloak instance and configuring

```yaml
authz:
  impl: kessel
  kessel:
    insecure-client: true
    url: localhost:9000
    enable-oidc-auth: true
    sa-client-id: "svc-test"
    sa-client-secret: "<secret>"
    sso-token-endpoint: "http://localhost:8084/realms/redhat-external/protocol/openid-connect/token"
```

Enabling authentication for relations API and configuring the same keycloak instance.

We don't have the `Kessel.check` or create tuple calls used much in the inventory-api. For testing I created some calls and removed them in host service.

jmelis (approved):
Abhishek and I have spoken, and we agree that there is no value in storing the token in the cache if we are also going to verify whether or not the token has expired for every request.

The plan is to merge this MR as is, but to remove the caching bit in a future MR (unless we find a good use case for the cache).

PR #93, opened 2024-09-05
RHCLOUD-34169 resource registration

jmelis:
## Describe your changes

- register the relation through authz between the resource and the workspace

Questions for reviewers:

- Does setting a default workspace make sense, or should the relationship be skipped in that scenario?

Fixes https://issues.redhat.com/browse/RHCLOUD-34169

## Checklist

### DIFF
--- a/data/host.json
+++ b/data/host.json
@@ -2,7 +2,7 @@
   "rhelHost": {
     "metadata": {
       "resource_type": "rhel-host",
-      "workspace": ""
+      "workspace": "workspace1"
     },
     "reporter_data": {
       "reporter_type": "OCM",
--- a/deploy/schema.yaml
+++ b/deploy/schema.yaml
@@ -85,3 +85,7 @@ schema: |-
     permission disable = workspace->notifications_integration_disable
     permission enable = workspace->notifications_integration_enable
   }
+
+  definition hbi/rhel_host {
+    relation workspace: rbac/workspace
+  }
--- a/internal/authz/allow/allow.go
+++ b/internal/authz/allow/allow.go
@@ -31,3 +31,7 @@ func (a *AllowAllAuthz) CreateTuples(ctx context.Context, r *kessel.CreateTuples
 func (a *AllowAllAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRequest) (*kessel.DeleteTuplesResponse, error) {
 	return &kessel.DeleteTuplesResponse{}, nil
 }
+
+func (a *AllowAllAuthz) SetWorkspace(ctx context.Context, id int64, workspace, name, namespace string) (*kessel.CreateTuplesResponse, error) {
+	return &kessel.CreateTuplesResponse{}, nil
+}
--- a/internal/authz/allow/allow.go
+++ b/internal/authz/allow/allow.go
@@ -32,6 +32,6 @@ func (a *AllowAllAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuples
 	return &kessel.DeleteTuplesResponse{}, nil
 }
 
-func (a *AllowAllAuthz) SetWorkspace(ctx context.Context, id int64, workspace, name, namespace string) (*kessel.CreateTuplesResponse, error) {
+func (a *AllowAllAuthz) SetWorkspace(ctx context.Context, local_resource_id, workspace, name, namespace string) (*kessel.CreateTuplesResponse, error) {
 	return &kessel.CreateTuplesResponse{}, nil
 }
--- a/internal/authz/kessel/kessel.go
+++ b/internal/authz/kessel/kessel.go
@@ -2,6 +2,9 @@ package kessel
 
 import (
 	"context"
+	"fmt"
+	"strconv"
+
 	"github.com/go-kratos/kratos/v2/log"
 	authzapi "github.com/project-kessel/inventory-api/internal/authz/api"
 	kessel "github.com/project-kessel/relations-api/api/kessel/relations/v1beta1"
@@ -68,4 +71,42 @@ func (a *KesselAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRe
 		return nil, err
 	}
 	return a.TupleService.DeleteTuples(ctx, r, opts...)
+}
+
+func (a *KesselAuthz) SetWorkspace(ctx context.Context, id int64, workspace, namespace, name string) (*kessel.CreateTuplesResponse, error) {
+	if workspace == "" {
+		return nil, fmt.Errorf("workspace is required")
+	}
+
+	// TODO: remove previous tuple for workspace
+
+	rels := []*kessel.Relationship{{
+		Resource: &kessel.ObjectReference{
+			Type: &kessel.ObjectType{
+				Name:      name,
+				Namespace: namespace,
+			},
+			Id: strconv.FormatInt(id, 10),
+		},
+		Relation: "workspace",
+		Subject: &kessel.SubjectReference{
+			Subject: &kessel.ObjectReference{
+				Type: &kessel.ObjectType{
+					Name:      "workspace",
+					Namespace: "rbac",
+				},
+				Id: workspace,
+			},
+		},
+	}}
+
+	response, err := a.CreateTuples(ctx, &kessel.CreateTuplesRequest{
+		Tuples: rels,
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return response, nil
 }
--- a/internal/authz/kessel/kessel.go
+++ b/internal/authz/kessel/kessel.go
@@ -3,7 +3,6 @@ package kessel
 import (
 	"context"
 	"fmt"
-	"strconv"
 
 	"github.com/go-kratos/kratos/v2/log"
 	authzapi "github.com/project-kessel/inventory-api/internal/authz/api"
@@ -73,7 +72,7 @@ func (a *KesselAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRe
 	return a.TupleService.DeleteTuples(ctx, r, opts...)
 }
 
-func (a *KesselAuthz) SetWorkspace(ctx context.Context, id int64, workspace, namespace, name string) (*kessel.CreateTuplesResponse, error) {
+func (a *KesselAuthz) SetWorkspace(ctx context.Context, local_resource_id, workspace, namespace, name string) (*kessel.CreateTuplesResponse, error) {
 	if workspace == "" {
 		return nil, fmt.Errorf("workspace is required")
 	}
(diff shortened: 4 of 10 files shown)


internal/data/hosts/hosts.go
```diff
@@ -49,6 +56,38 @@ func (r *hostsRepo) Save(ctx context.Context, model *biz.Host) (*biz.Host, error
 			return nil, err
 		}
 	}
+
+	if r.Authz != nil {
+		rels := []*kessel.Relationship{{
+			Resource: &kessel.ObjectReference{
+				Type: &kessel.ObjectType{
```
Rajagopalan-Ranganathan, line 63: not sure of our future use cases - but the name and namespace are hardcoded here , would it be future safe to parametrize them? WDYT?

internal/data/hosts/hosts.go
```diff
@@ -33,6 +35,11 @@ func (r *hostsRepo) Save(ctx context.Context, model *biz.Host) (*biz.Host, error
 		return nil, nil
 	}
 
+	// Set a default value for workspace if it's not set
+	if model.Metadata.Workspace == "" {
+		model.Metadata.Workspace = "default"
```
wscalf, line 40: It may just a placeholder, but a note: object ids in Zanzibar are global, so there would only be one "default" workspace shared by all orgs, which probably isn't intended.

internal/data/hosts/hosts.go
jmelis: This is a good point that is also causing me a bit of concern. Essentially we are tying the implementation to the specific internal schema we are using. As Alec put it this is accepted technical debt that we don't need to resolve right now.

I see it like you do, parametrizable in the future or maybe plugin driven.

internal/data/hosts/hosts.go
jmelis: Right, it's not a placeholder, I was proposing to merge it as is. The other option is to not set the relationship if the workspace is not passed.

What do you recommend?

cc @alechenninger

alechenninger: So for hosted we need this to pull from the user's auth method. Basically, the default workspace for a resource needs to be based on the user's org ID.

I imagine something like this:

- As I understand there is some authentication abstraction in Inventory, the output of this should have, among other things, a notion of the user's "tenant" or "organization." ("Tenant" is what is used in RBAC... I would reuse that for consistency since that will be moved under project-kessel as well)
- At some point then in processing the create (probably outside of the repository?), if a workspace is not provided the default should be derived from the user's tenant.
- The current logic we have in RBAC is that the default workspace ID is exactly equal to the user's org ID. This may change before production but for now this is a fine starting point I think (and both RBAC and Inventory will agree, which is an important requirement)

Does this makes sense?

internal/data/hosts/hosts.go
alechenninger: For a host repository this should be something like "hbi" namespace and "host" type. We need an "integration repository" for integrations I think, which I think we have: https://github.com/project-kessel/inventory-api/blob/main/internal/data/notificationsintegrations/notificationsintegrations.go

The confusion of hosts vs integrations is my fault – we were originally prioritizing Hosts from HBI in order to get access aware notifications for hosts, but then pivoted for various reasons to start with based RBAC for Notifications integrations as a more minimal starting point that involved less teams.

internal/data/hosts/hosts.go
```diff
@@ -49,6 +56,38 @@ func (r *hostsRepo) Save(ctx context.Context, model *biz.Host) (*biz.Host, error
...
+				Id: strconv.FormatInt(model.ID, 10),
+			},
+			Relation: "workspace",
+			Subject: &kessel.SubjectReference{
+				Subject: &kessel.ObjectReference{
+					Type: &kessel.ObjectType{
+						Name:      "workspace",
+						Namespace: "rbac",
+					},
+					Id: model.Metadata.Workspace,
```
josejulio, line 77: Do we need to check if the caller has access to the workspace? or this check is assumed in this call?

alechenninger: There should be checks that the reporter is allowed to report for this resource type. That won't be workspace specific.

Workspace would be relevant for the end user. End use access checks have come up a few times in the last few weeks. I have a proposal that we should but we can leave it off of this PR. I'll share that with you tomorrow.

internal/data/hosts/hosts.go
randymgeorge: hosts has been changed to rhel_hosts.  Not sure when Chris made that change. However, some of the framework is still "hosts".  When you say "namespace", are you referring a reporter?  HBI is a reporter

internal/data/hosts/hosts.go
randymgeorge: @alechenninger Currently, inventory does not capture the org id.  Question, what if the reporter, e.g. HBI, is multi-tenant. We can not pull the org ID from the authN of the reporter to get the org ID for the resource.

deploy/schema.yaml
```diff
@@ -85,3 +85,7 @@ schema: |-
     permission disable = workspace->notifications_integration_disable
     permission enable = workspace->notifications_integration_enable
   }
+
+  definition hbi/rhel_host {
```
josejulio, line 89: We are adopting the resource type name for the schema `rhe_host`

Should this change go over to https://github.com/RedHatInsights/kessel-config/blob/master/schema.zed instead ?

josejulio:
Tried it locally using `make inventory-up`

Send the sample request without a workspace and got:
```
{"code":500,"reason":"","message":"workspace is required","metadata":{}}
```

Relations API didn't show any entry (i.e. wasn't hit).

After using a workspace (`someday`) I got a 200 response and relations API shows the following:

```
INFO ts=2024-09-10T18:28:18Z caller=log/helper.go:80 service.id=relations-api service.name= service.version= trace.id= span.id= kind=server component=grpc operation=/kessel.relations.v1beta1.KesselTupleService/CreateTuples args=tuples:{resource:{type:{namespace:"hbi"  name:"rhel_host"}  id:"2"}  relation:"workspace"  subject:{subject:{type:{namespace:"rbac"  name:"workspace"}  id:"someday"}}} code=0 reason= stack= latency=0.131146415
```

josejulio:
@alechenninger Could you gives us a review on this approach?

We are only considering that workspace is required if we are connecting to Kessel, otherwise it is not used at all.

Thanks

internal/data/hosts/hosts.go
jmelis: for now, we have agreed to not do any checks

deploy/schema.yaml
jmelis: it should, @alechenninger is working on that

akoserwal (approved):

deploy/schema.yaml
randymgeorge: There is a default workspace.  It should not be required.  We are going in circles here due to the lack of a design which is reviewed and approved.  This PR should be blocked on that design.  @jewzaam will provide this design

internal/authz/allow/allow.go
```diff
@@ -31,3 +31,7 @@ func (a *AllowAllAuthz) CreateTuples(ctx context.Context, r *kessel.CreateTuples
 func (a *AllowAllAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRequest) (*kessel.DeleteTuplesResponse, error) {
 	return &kessel.DeleteTuplesResponse{}, nil
 }
+
+func (a *AllowAllAuthz) SetWorkspace(ctx context.Context, id int64, workspace, name, namespace string) (*kessel.CreateTuplesResponse, error) {
```
alechenninger, line 35: I don't think we want to require the ID of the resource to be `int64` right?

We need the reporter's ID to be stored in relations graph, in addition to the inventory ID. As the inventory ID has no persistence for it today otherwise, it's not needed right now.

So in order to support the reporter's ID, which I don't believe we need to (or should) constrain, I think we want this to be an arbitrary string (essentially only limited by the resource ID constraints of the relations API which is very minimal, essentially no whitespace and a few other chars IIRC).

internal/authz/kessel/kessel.go
```diff
@@ -69,3 +72,41 @@ func (a *KesselAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRe
 	}
 	return a.TupleService.DeleteTuples(ctx, r, opts...)
 }
+
+func (a *KesselAuthz) SetWorkspace(ctx context.Context, id int64, workspace, namespace, name string) (*kessel.CreateTuplesResponse, error) {
+	if workspace == "" {
+		return nil, fmt.Errorf("workspace is required")
+	}
+
+	// TODO: remove previous tuple for workspace
```
alechenninger, line 81: Yes this needs to be done or else a resource would be in multiple workspaces which is a constraint violation.

However, this needs to be done in one atomic call which the relations API doesn't support yet. I'll file a ticket for resolving this and the relations API additions necessary.

And as of now as we know, there is no workspace movement needed, so omitting this should suffice for the next few weeks or so.

internal/data/notificationsintegrations/notificationsintegrations.go
```diff
@@ -49,6 +49,14 @@ func (r *notificationsintegrationsRepo) Save(ctx context.Context, model *biz.Not
 			return nil, err
 		}
 	}
+
+	if r.Authz != nil {
+		_, err := r.Authz.SetWorkspace(ctx, model.ID, model.Metadata.Workspace, "notifications", "integration")
```
alechenninger, line 54: Which ID is this? I think this may be wrong.

We need this to be the _reporter's_ ID–as in, Notification's ID for the Integration (or hbi's, or acm's, etc).

This is because for queries and access control, these systems will be using their own IDs, not Kessel's IDs. So for initial use cases to work especially, these have to be reporter IDs for the resource.

alechenninger (changes requested):
Main change I think is that we need to use the Reporter's ID for the resource. Let me know if there are any questions!

If it's possible to accommodate this one change before moving on that would be great.

jmelis:
@alechenninger that is a massive oversight from our side! thank you so much for catching it 🤦

josejulio:
I just noticed that we didn't add the same logic to `k8spolicies`

https://github.com/project-kessel/inventory-api/blob/cba4fbc81592d68c23e586a6ecec34250145ae15/internal/data/k8spolicies/k8spolicies.go#L52-L51

data/host.json
```diff
@@ -2,7 +2,7 @@
   "rhelHost": {
     "metadata": {
       "resource_type": "rhel-host",
-      "workspace": ""
+      "workspace": "workspace1"
```
josejulio, line 5: Could you also update the [k8s-cluster](https://github.com/project-kessel/inventory-api/blob/0be975b3336ba576b96ef8b6ababe27e7d089759/data/k8s-cluster.json#L5) example?

jmelis:
> I just noticed that we didn't add the same logic to `k8spolicies`
> 
> https://github.com/project-kessel/inventory-api/blob/cba4fbc81592d68c23e586a6ecec34250145ae15/internal/data/k8spolicies/k8spolicies.go#L52-L51

good catch!

data/host.json
jmelis: I intentionally did not update it because that's for on-prem right now, which will use allow-all, and it may cause more confusion if we do?

josejulio: oh, that makes sense

josejulio (approved):

josejulio (approved):
lgtm

PR #168, opened 2024-09-26
RHCLOUD-35379 updates image tag variable to leverage app interface better

tonytheleg:
## Describe your changes

* App interface, when using openshift templates, makes the `IMAGE_TAG` variable available in all deployments for setting the image tag. This value is generated from the short hash of the git commit which matches our image tags in Quay. This will ensure the desired commit is always running versus using `latest` which has security issues and may also not be latest with regards to code changes

See [DOCS](https://gitlab.cee.redhat.com/service/app-interface/-/blob/master/docs/app-sre/continuous-delivery-in-app-interface.md#automatically-generated-parameters)

### DIFF
--- a/deploy/kessel-inventory.yaml
+++ b/deploy/kessel-inventory.yaml
@@ -53,11 +53,11 @@ objects:
           podSpec:
             initContainers:
             - name: migration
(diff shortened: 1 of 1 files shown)


josejulio (approved):
lgtm

PR #196, opened 2024-10-16
Added metrics for relation-api calls via inventory-api

Adam0Brien:
## Describe your changes

- Adding success and failure metrics for relations-api calls that are made through inventory-api
- Bumped go version to `1.23.1` because of a [security vulnerability ](https://github.com/project-kessel/inventory-api/actions/runs/11369611074)

Fixes #

## Checklist


internal/authz/kessel/kessel.go
```diff
@@ -9,49 +9,85 @@ import (
...
+		tokenClient:    tokenCli,
+		successCounter: successCounter,
+		failureCounter: failureCounter,
+	}, nil
 }
 
 func (a *KesselAuthz) Health(ctx context.Context) (*kesselv1.GetReadyzResponse, error) {
 	opts, err := a.getCallOptions()
 	if err != nil {
+		a.failureCounter.Add(context.Background(), 1, metric.WithAttributes(attribute.String("method", "Health")))
```
tonytheleg, line 60: Possibly nitpicking or may not provide much value now, but since we have to frequently make calls to these counters, would it be good to make this a method of the KesselAuthz type to simplify?

```go
func (a *KesselAuthz) incrFailureCounter(method string) {
    a.failureCounter.Add(context.Background(), 1, metric.WithAttributes(attribute.String("method", method)))
}

func (a *KesselAuthz) incrSuccessCounter(method string) {
    a.successCounter.Add(context.Background(), 1, metric.WithAttributes(attribute.String("method", method)))
}
```

Then a failure or success would be
```go
a.incrFailureCounter("Health")
a.incrSuccessCounter("CreateTuple")
```
etc, etc

It may seem pointless now, but if we ever need to add more attributes other than just method, we can update it in 2 places instead of potentially 20. Thinking it might make it easier down the road if needed kinda deal. Plus if we need to make better use of the context for any reason and want to add a context to the KesselAuthz type to pass around it also makes it an easy change. Either way, this looks great to me!

Adam0Brien: Makes sense to me, Ill include this in the next push!

tonytheleg (approved):

akoserwal (approved):

PR #201, opened 2024-10-21
fixing format for principal identity to be domain/sub

akoserwal:
## Describe your changes

Updating the principal identity format from domain: sub to "domain (without ".com"/sub)

- ...

Fixes #

## Checklist

### DIFF
--- a/.inventory-api.yaml
+++ b/.inventory-api.yaml
@@ -6,6 +6,11 @@ server:
     address: localhost:9000
 authn:
    allow-unauthenticated: true
+#   oidc:
+#     authn-server-url: http://localhost:8084/realms/redhat-external
+#     skip-client-id-check: true
+#     skip-issuer-check: true
+#     principal-user-domain: localhost
 authz:
   impl: allow-all
 eventing:
--- a/internal/authn/oidc/oidc.go
+++ b/internal/authn/oidc/oidc.go
@@ -4,10 +4,8 @@ package oidc
 import (
 	"context"
 	"fmt"
-	"regexp"
-	"strings"
-
 	coreosoidc "github.com/coreos/go-oidc/v3/oidc"
+	"github.com/go-kratos/kratos/v2/log"
 	"github.com/go-kratos/kratos/v2/transport"
 
 	"github.com/project-kessel/inventory-api/internal/authn/api"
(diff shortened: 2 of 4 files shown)


.inventory-api.yaml
```diff
@@ -6,6 +6,11 @@ server:
     address: localhost:9000
 authn:
    allow-unauthenticated: true
+#   oidc:
+#     authn-server-url: http://localhost:8084/realms/redhat-external
+#     skip-client-id-check: true
+#     skip-issuer-check: true
+#     principal-user-domain: localhost
```
josejulio, line 13: Is this comited on purpose? should we remove if it is no longer relevant ?

internal/authn/oidc/oidc.go
```diff
@@ -97,23 +96,27 @@ func issuerCheck(issuer string, domain string) bool {
...
+		issuer = issuer[:dotIndex]
+	}
+
+	if strings.HasSuffix(issuer, "."+domain) {
 		return true
 	}
-	return false
+
+	return issuer == domain
 }
```
josejulio, line 122: Could we add some tests against this issuerCheck?

internal/authn/oidc/oidc.go
```diff
@@ -97,23 +96,27 @@ func issuerCheck(issuer string, domain string) bool {
...
+	if dotIndex := strings.LastIndex(issuer, "."); dotIndex != -1 {
+		issuer = issuer[:dotIndex]
+	}
+
+	if strings.HasSuffix(issuer, "."+domain) {
 		return true
 	}
-	return false
+
+	return issuer == domain
```
alechenninger, line 121: So the domain is not _literally_ the DNS domain of the issuer – it is just an arbitrary/abstract "domain" identifier (as in, "trust domain"). So what I would do is when configuring an issuer, have the "domain" that is used be configurable per that issuer. You have the "principal-user-domain" config already – basically use that as long as the token validates with a key from that issuer.

.inventory-api.yaml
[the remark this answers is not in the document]
akoserwal: This file used for the default run option for inventory API locally,  where I added oidc config comment as an example. If someone wants to test with sso integration locally

akoserwal: Yes, I add on purpose. For devs who want to run a local keycloak and test oidc flow

internal/authn/oidc/oidc.go
akoserwal: remove the issuer check

internal/authn/oidc/oidc.go
akoserwal: fixed

josejulio (approved):

PR #212, opened 2024-10-30
improves error messages for better error logging/troubleshooting

tonytheleg:
## Describe your changes

The goal of this PR is to update any error messages/logs where the issue may not be clear what failed. That includes:
* Methods that have multiple error checks -- each error returned now at least explains what failed
* Methods that don't return errors but have error checks, errors are now logged to know what failed
* Any instantiation that may call external packages where the error is clear but the reason its being called is not.

These should all loop back to a logger as it bubbles up, which will ensure any error logs can be easily correlated to some code logic hopefully and make troubleshooting easier

For RHCLOUD-35835

## Checklist

### DIFF
--- a/internal/authn/oidc/oidc.go
+++ b/internal/authn/oidc/oidc.go
@@ -62,11 +62,13 @@ func (o *OAuth2Authenticator) Authenticate(ctx context.Context, t transport.Tran
 	u := &Claims{}
 	err = tok.Claims(u)
 	if err != nil {
+		log.Errorf("failed to extract claims: %v", err)
 		return nil, api.Deny
 	}
 
 	if o.EnforceAudCheck {
 		if u.Audience != o.CompletedConfig.ClientId {
+			log.Debugf("aud does not match the requesting client-id -- decision DENY")
 			return nil, api.Deny
 		}
 	}
--- a/internal/authn/psk/config.go
+++ b/internal/authn/psk/config.go
@@ -2,10 +2,9 @@ package psk
 
 import (
 	"fmt"
+	"gopkg.in/yaml.v3"
 	"io"
 	"os"
-
-	"gopkg.in/yaml.v3"
 )
 
 type Config struct {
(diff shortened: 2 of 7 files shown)


internal/authn/oidc/oidc.go
```diff
@@ -62,11 +62,13 @@ func (o *OAuth2Authenticator) Authenticate(ctx context.Context, t transport.Tran
 	u := &Claims{}
 	err = tok.Claims(u)
 	if err != nil {
+		log.Errorf("failed to extract claims: %v", err)
 		return nil, api.Deny
 	}
 
 	if o.EnforceAudCheck {
 		if u.Audience != o.CompletedConfig.ClientId {
+			log.Warnf("aud does not match the requesting client-id -- decision DENY")
```
josejulio, line 71: I wonder if those should be info or even debug :thinking:

tonytheleg: Yeah this one I wasn't 100% on. if you have `enforce-aud-check` on and this check fails, all youre going to get is an unauthorized error, it would be nice to know why it was unauthorized when it fails in this particular place. I didnt necessarily want to call it an error, but if its INFO, its not going to jump out in the logs very well with the hundreds of other INFO logs, WARN i thought was a good middle ground? I'm pretty flexible if we think it should be something else, but just providing my thinking if it helps.

josejulio: My only concern is that this is a problem with the user input, not a real error with the application itself or its configuration. It will correctly deny the access if the user is holding the wrong credentials / token.

tonytheleg: Fair enough, I'm thinking debug then as my main point for wanting the log here is so that the denial isn't hidden while testing if it's an aud check. If it's really only for testing debug is more reasonable. Sound good?

josejulio: It sounds good to me

tonytheleg: updated to use debug instead of warn.

josejulio (approved):
lgtm

PR #220, opened 2024-11-05
RHCLOUD-35930: converts uint64 id's to uuidV7

tonytheleg:
## Describe your changes

NOTE: Merging this PR before wiping out DB's to start fresh will fail. Details in card.

* Converts all ID's/Resource ID's from uint64 to uuid v7
* Converts eventing metadata ID's to strings vs uint64 as its easier to get the string value of uuid vs int
  * There didnt seem to be an apparent benefit/reason to use UUID's in events for ID's, but if there is one, open to changing them
* Updates all existing tests to use uuid's
* Adds logic to generate the UUID on Create calls

Validation:
I've tested this pretty extensively in ephemeral and captured database outputs from testing in a file added to the jira ticket. See [LINK](https://issues.redhat.com/browse/RHCLOUD-35930?focusedId=26010700&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26010700_ for more details

Fixes #

## Checklist

### DIFF
--- a/internal/biz/model/relationships.go
+++ b/internal/biz/model/relationships.go
@@ -3,18 +3,20 @@ package model
 import (
 	"database/sql/driver"
 	"encoding/json"
+	"fmt"
+	"github.com/google/uuid"
 	"gorm.io/gorm"
 	"gorm.io/gorm/schema"
 	"time"
 )
 
 type Relationship struct {
-	ID               uint64 `gorm:"primarykey"`
-	OrgId            string `gorm:"index"`
+	ID               uuid.UUID `gorm:"type:uuid;primarykey"`
+	OrgId            string    `gorm:"index"`
 	RelationshipData JsonObject
 	RelationshipType string
-	SubjectId        uint64 `gorm:"index;not null"`
-	ObjectId         uint64 `gorm:"index;not null"`
+	SubjectId        uuid.UUID `gorm:"type:uuid;index;not null"`
+	ObjectId         uuid.UUID `gorm:"type:uuid;index;not null"`
 	Reporter         RelationshipReporter
 	CreatedAt        *time.Time
 	UpdatedAt        *time.Time
@@ -42,4 +44,15 @@ func (data RelationshipReporter) Value() (driver.Value, error) {
 
 func (data *RelationshipReporter) Scan(value interface{}) error {
 	return Scan(value, data)
+}
+
+func (r *Relationship) BeforeCreate(db *gorm.DB) error {
+	var err error
+	if r.ID == uuid.Nil {
+		r.ID, err = uuid.NewV7()
+		if err != nil {
+			return fmt.Errorf("failed to generate resource uuid: %w", err)
+		}
+	}
+	return nil
 }
--- a/internal/biz/model/resourcehistory.go
+++ b/internal/biz/model/resourcehistory.go
@@ -2,14 +2,15 @@ package model
 
 import (
 	"fmt"
+	"github.com/google/uuid"
 	"gorm.io/gorm"
 	"gorm.io/gorm/schema"
 	"time"
 )
 
 type ResourceHistory struct {
-	ID           uint64 `gorm:"primarykey"`
-	OrgId        string `gorm:"index"`
+	ID           uuid.UUID `gorm:"type:uuid;primarykey"`
+	OrgId        string    `gorm:"index"`
 	ResourceData JsonObject
 	ResourceType string
 	WorkspaceId  string
@@ -19,7 +20,7 @@ type ResourceHistory struct {
 	Labels       Labels
 	Timestamp    *time.Time `gorm:"autoCreateTime"`
 
-	ResourceId    uint64        `gorm:"index"`
+	ResourceId    uuid.UUID     `gorm:"type:uuid;index"`
 	OperationType OperationType `gorm:"index"`
 }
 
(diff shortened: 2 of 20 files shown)


internal/biz/model/relationships.go
```diff
@@ -43,3 +45,28 @@ func (data RelationshipReporter) Value() (driver.Value, error) {
...
+func (r *Relationship) BeforeCreate(db *gorm.DB) error {
+	var err error
+	if r.ID == uuid.Nil {
+		r.ID, err = uuid.NewV7()
+		if err != nil {
+			return fmt.Errorf("failed to generate resource uuid: %w", err)
+		}
+	}
+
+	if r.SubjectId == uuid.Nil {
```
tylercreller, line 58: Based on my super naive understanding of relationships I believe subject and object should already exist before the relationship is able to be created, is that correct? If so, we can definitely remove the SubjectId and ObjectId fallback uuid creation here since we won't need to ever create these when a relationship is created.

internal/biz/model/resourcehistory.go
```diff
@@ -40,3 +41,21 @@ func (r *ResourceHistory) ResourceHistory(db *gorm.DB, s *schema.Schema) error {
...
+func (r *ResourceHistory) BeforeCreate(db *gorm.DB) error {
+	var err error
+	if r.ID == uuid.Nil {
+		r.ID, err = uuid.NewV7()
+		if err != nil {
+			return fmt.Errorf("failed to generate uuid: %w", err)
+		}
+	}
+
+	if r.ResourceId == uuid.Nil {
```
tylercreller, line 54: Same thing for resource here as my other comment, I believe the resource should already exist and have an ID, and if not we may have a bigger issue.

internal/biz/model/relationships.go
josejulio: That's right - these should exist already. 
+1 on removing the SubjectId and ObjectId checks

internal/biz/model/resourcehistory.go
josejulio: +1

internal/biz/model/relationships.go
tonytheleg: my also naive understanding here; subject ID and object ID in this scenario are unique inside of the inventory DB are are not configurable/exposed via API. 

When a user creates a relationship they DO have to provide a subject_local_resource_id and object_local_resource_id which are set in the reporter_data section of the request/table. My understanding is those local_resource_id values are supposed to be some sort of ID that is unique to the requesting system (think OCM clusters to provision shards or whatever values)

The subject_id and object_id values were made UUID to improve queries in inventory as they are used to find a relationship based on https://github.com/project-kessel/inventory-api/blob/a85a24d48f9b4ce3130a4f06c7d6dd8cc12e82d0/internal/data/relationships/relationships.go#L105

https://github.com/project-kessel/docs/blob/main/src/content/docs/inventory/inventory-api.md#reporter_data-1

tonytheleg: sample relationship in DB where subjectID and objectID are UUID and the provided local resource ids show in reporter section

```kessel-inventory=# select * from relationships;
-[ RECORD 1 ]-----+----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
id                | 0192fe0f-3520-7e45-bf91-e6396d872872
org_id            | 
relationship_data | {"status": 3}
relationship_type | k8s-policy_is-propagated-to_k8s-cluster
subject_id        | 0192fe0f-095f-7be6-a25a-a15eab7ffc20
object_id         | 0192fe0e-c7e9-7cb1-9c12-02bc105d5b40
reporter          | {"reporter_id": "curl/8.6.0", "reporter_type": "OCM", "reporter_version": "0.1", "object_resource_type": "k8s-cluster", "subject_resource_type": "k8s-policy", "object_local_resource_id": "2", "subject_local_resource_id": "1"}
created_at        | 
updated_at        | 2024-11-05 20:41:15.727725+00
```

internal/biz/model/resourcehistory.go
tonytheleg: in this scenario, i think you are right. resource ID should be the UUID created for a resource (ID) and you can't have history on a resource without having a resource ID. Ill update this and give it a test, good call!

internal/biz/model/relationships.go
josejulio: In the case of relationships, when we are creating one, we do get a subject_local_resource_id and object_local_resource_id.
These values (along with the types and resource data) should map to `resource` (which already have an id) - we take the id from there and populate or model.

See
https://github.com/project-kessel/inventory-api/blob/main/internal/biz/relationships/relationships.go#L48-L66

tonytheleg: I'm following now. Just updated the logic and tested it and it makes perfect sense. Good eyes @tylercreller !

tonytheleg:
Updated to remove setting Subject/Object ID and Resource ID in history as these should correlate to existing resource ID's

Thanks @josejulio and @tylercreller for the info, makes total sense after running it without the changes

tylercreller:
Might be able to use dependabot's PR for the image bump https://github.com/project-kessel/inventory-api/pull/221/files

tonytheleg:
> Might be able to use dependabot's PR for the image bump https://github.com/project-kessel/inventory-api/pull/221/files

Didnt even notice the dependabot PR's were setup. +1

tylercreller:
LGTM

PR #256, opened 2024-11-21
update deployment to ensure RDS SSL conction and CA verification

tonytheleg:
## Describe your changes

* updates config package to
  * add logic to check if there is an RDS CA cert value in cdapp config -- if so, it sets ssl mode to verify and leverages the RdsCa method from app-common to dump the cert to a temp file for consumption
  * add error handling for above changes
* updates test for changes to ensure the RDS CA value is properly set and CA contents match app config value

Validation:
Can't really validate this in ephemeral easily without an RDS db provided by App Interface due to how ClowdApps work. 

The tests confirm the logic works if the data is there, but we would need to fully validate this functionality in stage since we can't really mock App interface or test in ephemeral with it.

Fixes #

## Checklist

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -76,7 +76,10 @@ func init() {
 	}
 
 	if clowder.IsClowderEnabled() {
-		options.InjectClowdAppConfig()
+		err := options.InjectClowdAppConfig()
+		if err != nil {
+			panic(err)
+		}
 	}
 
 	loggerOptions := common.LoggerOptions{
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"fmt"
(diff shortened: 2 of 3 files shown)


tonytheleg:
Found out that the [app-common](https://github.com/RedHatInsights/app-common-go) pkg we're already using has a helper for setting up an RDS CA file for consumption

>The clowder library also comes with several other helpers
>    clowder.LoadedConfig.RdsCa() - creates a temporary file with the RDSCa and returns the filename.

Going to look into that option as it may not require any changes in deployment, just in config logic

tylercreller:
LGTM - no comments. If you haven't, consider also testing things work fine with persistence disabled for good measure.

Adam0Brien (approved):
LGTM

PR #293, opened 2025-01-15
RHCLOUD-37203: updates build and code to add FIPS support

tonytheleg:
## Describe your changes

**Adds `fips.go`:**
* new package to restrict all TLS configuration to FIPS-approved settings (See [fipsonly.go](https://go.dev/src/crypto/tls/fipsonly/fipsonly.go))

**Rolls back go version:**
* `go-toolset` currently installs go 1.22.9, so `go.mod` has been updated to ensure a supported version of Go is used for builds to work

**Updates to Dockerfile:**
* adds the `go-toolset` package to both  the builder container and the final container versus installing Go from upstream
  * the version of Go from `go-toolset` contains modifications made for RHEL which replaces BoringSSL with OpenSSL. OpenSSL is approved and validated for FedRAMP and FIPS, BoringSSL is not
  * making `go tool` available on the final image will also aid in proving FIPS compliance for audits
* installs `fips-detect` to simplify proving FIPS validation during audits as well (More on [fips-detect](https://github.com/acardace/fips-detect)

**Updates to Makefile:**
* adds a `FIPS_ENABLED` var that defaults to true
* adds some GO env vars to be more explicit for builds, including setting GOBUILDFLAGS to remove paths with compiler and assembler (cleaner errors for things like panic's by stripping full paths to packages)
* adds conditional to enforce FIPS during build if FIPS_ENABLED=true
* updates build command to use new buildflags and account for FIPS or not
* adds `local-build` target since building with FIPS_ENABLED will fail locally and wanted to make it easier to build locally without having to set FIPS_ENABLED=false

Updates to README:
* adds info on building locally 
* adds notes on testing the API in stage and how to test creating/deleting notifications-integrations which work in stage
* adds info on how to validate FIPS for audits/any other reason

Some notes/useful links:
* many of the FIPS changes to build flags, fips.go etc are well tested in SREP for ensuring all operators deployed for managed openshift are FIPS compliant as they run in FedRAMP. Some info on those changes are available [HERE](https://github.com/openshift/boilerplate/tree/master/boilerplate/openshift/golang-osd-operator#fips-federal-information-processing-standards) and can also be found in the Makefile in the same repo path
* [Is your Go Application FIPS Compliant?](https://developers.redhat.com/articles/2022/05/31/your-go-application-fips-compliant)
* long term we probably dont want to install go onto the running image, or install fips-detect at all but for the purpose of the FedRAMP build out and SCR ive added them now to simplify testing. We can remove them at a later day and come up with a better validation proof method

For RHCLOUD-37203

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,12 +2,7 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1154 AS builder
 
 ARG TARGETARCH
 USER root
-RUN microdnf install -y tar gzip make which gcc gcc-c++ cyrus-sasl-lib findutils git
-
-# install platform specific go version
-RUN curl -O -J  https://dl.google.com/go/go1.22.5.linux-${TARGETARCH}.tar.gz
-RUN tar -C /usr/local -xzf go1.22.5.linux-${TARGETARCH}.tar.gz
-RUN ln -s /usr/local/go/bin/go /usr/local/bin/go
+RUN microdnf install -y tar gzip make which gcc gcc-c++ cyrus-sasl-lib findutils git go-toolset
 
 WORKDIR /workspace
 
@@ -24,9 +19,16 @@ COPY main.go Makefile ./
 ARG VERSION
 RUN VERSION=${VERSION} make build
 
+# adds fips-detect tool for FIPS validation -- likely not needed long term
+RUN mkdir /tmp/go && GOPATH=/tmp/go GOCACHE=/tmp/go go install github.com/acardace/fips-detect@latest
+
 FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10-1154
 
+# installs RHEL fork of go to be able to validate with go tools for FIPS -- likely not needed long term
+RUN microdnf install -y go-toolset
+
 COPY --from=builder /workspace/bin/inventory-api /usr/local/bin/
+COPY --from=builder /tmp/go/bin/fips-detect /usr/local/bin/
 
 EXPOSE 8081
 EXPOSE 9081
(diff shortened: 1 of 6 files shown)


tylercreller:
Looks reasonable to me, I do have concerns that downgrading go might back us into a corner in the future but we can deal with that if it comes.

tonytheleg:
> Looks reasonable to me, I do have concerns that downgrading go might back us into a corner in the future but we can deal with that if it comes.

Same here, hopefully we won't hit that issue often if at all but we'll cross that bridge when it comes. I haven't been able to find a release cadence for go toolset yet, if we know how far behind it is from go (go 1.23 release august and looks like a 6 month cycle) it may not be a huge deal.

tonytheleg:
There also appears to be a go-toolset base image if ever needed, but i imagine the release cycle will be the same.

tonytheleg:
One minor final tweak that moves the fips.go file under the cmd package to ensure it gets called on any commands. I did not see the expected init function print out in logs, now I do which ensures the fipsonly is getting set

```
$ oc logs kessel-inventory-api-55d879c4f4-lbxqp | head
Defaulted container "kessel-inventory-api" out of: kessel-inventory-api, caddy-tls, migration-init (init)
***** Starting with FIPS crypto enabled *****  <--- fips.go init function output
INFO msg=Using config file: /inventory/inventory-api-config.yaml
Log Level is set to: info
Log Level is set to: info
INFO ts=2025-01-16T17:30:15Z caller=log/log.go:30 service.name=inventory-api service.version=0.1.0 trace.id= span.id= subsystem=storage msg=Persistence disabled: true
INFO ts=2025-01-16T17:30:15Z caller=log/log.go:30 service.name=inventory-api service.version=0.1.0 trace.id= span.id= subsystem=storage msg=Persistence disabled, skipping database connection...
```

PR #328, opened 2025-02-04
Fix delete api

snehagunta:
## Describe your changes

- Currently the delete API has a TODO in main to forward the call to RELATIONS
- This PR implements that
- [Confirmation from the relations team ](https://redhat-internal.slack.com/archives/C059NUD49PD/p1738692896892169?thread_ts=1738687564.185699&cid=C059NUD49PD) re. the call that needs to be made 

Fixes #

## Checklist

### DIFF
--- a/internal/authz/kessel/kessel.go
+++ b/internal/authz/kessel/kessel.go
@@ -3,6 +3,7 @@ package kessel
 import (
 	"context"
 	"fmt"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/spf13/viper"
 
@@ -148,6 +149,19 @@ func (a *KesselAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRe
 
 	a.incrSuccessCounter("DeleteTuples")
 	return resp, nil
+}
+
+func (a *KesselAuthz) UnsetWorkspace(ctx context.Context, local_resource_id, namespace, name string) (*kessel.DeleteTuplesResponse, error) {
+
+	req := &kessel.RelationTupleFilter{
+		ResourceNamespace: proto.String(namespace),
+		ResourceType:      proto.String(name),
+		ResourceId:        proto.String(local_resource_id),
+		Relation:          proto.String("workspace"),
+	}
+	return a.DeleteTuples(ctx, &kessel.DeleteTuplesRequest{
+		Filter: req,
+	})
 }
 
 func (a *KesselAuthz) SetWorkspace(ctx context.Context, local_resource_id, workspace, namespace, name string) (*kessel.CreateTuplesResponse, error) {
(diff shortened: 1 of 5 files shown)


internal/biz/resources/resources.go
```diff
@@ -174,7 +174,12 @@ func (uc *Usecase) Delete(ctx context.Context, id model.ReporterResourceId) erro
 		}
 	}
 
-	// TODO: delete the workspace tuple
+	if uc.Authz != nil {
+		_, err := uc.Authz.UnsetWorkspace(ctx, id.LocalResourceId, uc.Namespace, id.ResourceType)
```
tylercreller, line 178: I believe we should be pointing at the internal biz function that was created

internal/authz/kessel/kessel.go
```diff
@@ -150,6 +151,19 @@ func (a *KesselAuthz) DeleteTuples(ctx context.Context, r *kessel.DeleteTuplesRe
...
 }
 
+func (a *KesselAuthz) UnsetWorkspace(ctx context.Context, local_resource_id, namespace, name string) (*kessel.DeleteTuplesResponse, error) {
+
+	req := &kessel.RelationTupleFilter{
+		ResourceNamespace: proto.String(namespace),
+		ResourceType:      proto.String(name),
+		ResourceId:        proto.String(local_resource_id),
+		Relation:          proto.String("workspace"),
+	}
```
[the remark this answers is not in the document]
tonytheleg: Better example since the above is not geared towards something Inventory would create
```json
{"filter":{"resource_id":"1234","resource_type":"integration","resource_namespace":"notifications","relation":"workspace","subject_filter":{"subject_type":"workspace","subject_namespace":"rbac","subject_id":"1234"}}}
```

internal/biz/resources/resources.go
tonytheleg: +1, i think something more like
```go
	if uc.Authz != nil {
		err := biz.DefaultUnsetWorkspace(ctx, id.LocalResourceId, uc.Namespace, id.ResourceType, uc.Authz)
		if err != nil {
			return err
		}
	}
```
to match other calls, seems like everything else calls DefaultWorkspace

snehagunta: I fixed this. Thank you for catching it!

internal/authz/kessel/kessel.go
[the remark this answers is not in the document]
snehagunta: I agree with you that is better to provide more input data points here for filtering out the tuple to delete. There are 2 things to consider though

1. we currently do not have access to the workspace Id via the input payload for the delete API and that is what we would provide in the subject filter. Since we have currently disabled persistence, we cannot lookup the workspace id in the DB. We could look it up in relations, but that would be similar to supplying this data to the delete method, since delete in Relations will internally do a lookup based on the filter parameters we provide

2. All input parameters in the delete payload are currently Optional and the only requirement is that atleast 1 is populated. So this should, technically work. I spoke to the Relations team about this [here](https://redhat-internal.slack.com/archives/C059NUD49PD/p1738687564185699)

Hopefully this makes sense?

PR #329, opened 2025-02-05
minor tweaks to scripts/docs for better podman support

tonytheleg:
## Describe your changes

* adds note to docs about podman and potential issues seen when using the compose make targets with only podman installed
* updates the build_deploy.sh script to account for docker or podman so that it can be leveraged by anyone locally, and still function with jenkins (pretty sure jenkins nodes no longer have docker anyway)

Fixes #

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,8 @@ export DOCKER=docker
 make api
 ```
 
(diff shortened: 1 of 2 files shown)


internal/biz/resources/resources.go
```diff
@@ -3,9 +3,10 @@ package resources
 import (
 	"context"
 	"errors"
-	"github.com/google/uuid"
 	"time"
 
+	"github.com/google/uuid"
```
akoserwal, line 8: Seems like an unwanted change?

akoserwal (approved):

internal/biz/resources/resources.go
tonytheleg: yeah im going to fix it this morning, its there difference between code editors i think and how VS Code go fmt wants to shuffle it down and separate the github packages from the standard library. Ill edit in VIM so it doesnt do that

PR #348, opened 2025-02-13
Update namespace and ID used for check requests

lennysgarage:
## Describe your changes

- `uc.Namespace` would refer to the `authz`, which is the namespace defined in `serve.go`. This would cause issues with requests trying to read from `authz/integration` for example.
- `resource.ID` refers to the model resource id in the DB, such as `01950038-ad75-7b8d-903e-6d35d65e763d` which won't find a corresponding object in spicedb. 

Fixes #

## Checklist


wscalf:
LGTM

* The kesselcheckservice is an unusual use of the resources usecase where it doesn't make sense to construct it for a specific namespace the way it does for a concrete resource API. Will be interesting to see how this evolves with dynamic resources as it's a similar usecase, but this makes sense for the PoC imo.
* Using localresourceid also makes sense for the PoC I think - it's what's currently being bound to the workspace here, though I suspect it's something to keep an eye on with the multi-reporter work.

PR #349, opened 2025-02-13
RHCLOUD-37390: multi-reporter PoC

tylercreller:
## Describe your changes

- Adds InventoryResource model to group resources by
  - An inventory resource will be created when a new resource is requested to be created without an inventory_id in the metadata of the request
  - Introduces inventory_id onto the metadata model to allow resources to be associated to existing inventory resources during creation
- Adds InventoryResourceId onto resources for grouping
- Introduces flattened reporter columns (ID, local resource ID) onto resources
- Updates all resource workspaces when another resource associated to the same inventory resource is updated
- Sends events/workspace updates for all impacted resources at creation and update time
- Utilizes gorm transactions to ensure atomicity during creation of resources
- Supports backwards compatibility with existing resources
- Fixes bug with created_at being nulled on an update

RHCLOUD-37390
RHCLOUD-38174


akoserwal (approved):

internal/biz/model/resources.go
```diff
@@ -4,24 +4,30 @@ import (
...
-	"time"
 )
 
 type Resource struct {
 	ID           uuid.UUID `gorm:"type:uuid;primarykey"`
+	InventoryId  uuid.UUID `gorm:"index"`
 	OrgId        string    `gorm:"index"`
 	ResourceData JsonObject
 	ResourceType string
 	WorkspaceId  string
```
snehagunta, line 20: Are you going to move this to Inventory Resource?

tylercreller: This field will now live on both the reporter resources and inventory resources. Since we plan to still publish events and workspace updates for the reporter resources, this field needs to remain on the resource model.

internal/biz/resources/resources.go
```diff
@@ -16,55 +18,85 @@ import (
...
 func (uc *Usecase) Create(ctx context.Context, m *model.Resource) (*model.Resource, error) {
 	ret := m // Default to returning the input model in case persistence is disabled
+	updatedResources := []*model.Resource{}
 
 	if !uc.DisablePersistence {
 		// check if the resource already exists
-		resource, err := uc.repository.FindByReporterResourceId(ctx, model.ReporterResourceIdFromResource(m))
+		existingResource, err := uc.repository.FindByReporterData(ctx, m.ReporterId, m.ReporterResourceId)
+		if err != nil && errors.Is(err, gorm.ErrRecordNotFound) {
+			// Deprecated: fallback case for backwards compatibility
```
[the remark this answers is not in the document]
snehagunta: we talked about this offline. Resolving this thread.

internal/biz/resources/resources.go
```diff
@@ -16,55 +18,85 @@ import (
...
-func New(repository ResourceRepository, authz authzapi.Authorizer, eventer eventingapi.Manager, namespace string, logger log.Logger, disablePersistence bool) *Usecase {
+func New(repository ResourceRepository, inventoryResourceRepository InventoryResourceRepository, authz authzapi.Authorizer, eventer eventingapi.Manager, namespace string, logger log.Logger, disablePersistence bool) *Usecase {
 	return &Usecase{
-		repository:         repository,
-		Authz:              authz,
-		Eventer:            eventer,
-		Namespace:          namespace,
-		log:                log.NewHelper(logger),
-		DisablePersistence: disablePersistence,
+		repository:                  repository,
```
snehagunta, line 55: What do you think of renaming this to ReporterResourceRespository?

internal/biz/resources/resources.go
```diff
@@ -16,55 +18,85 @@ import (
...
 		}
-		if resource != nil {
+
+		if existingResource != nil {
 			return nil, ErrResourceAlreadyExists
 		}
 
-		ret, err = uc.repository.Save(ctx, m)
+		// Inventory Resource Validations
+		if m.InventoryId != nil {
```
snehagunta, line 86: This is where I was concerned about throwing an exception if we do not find a record for a given InventoryResourceId, but now I understand that flow will never enter this block since we are setting InventoryId  to uuid.Nil earlier in the flow.

internal/biz/resources/resources.go
```diff
@@ -119,21 +186,24 @@ func (uc *Usecase) Update(ctx context.Context, m *model.Resource, id model.Repor
...
+			}
 		}
 	}
 
 	if uc.Authz != nil {
-		// Todo: Update workspace if there is any change
-		err := biz.DefaultSetWorkspace(ctx, uc.Namespace, m, uc.Authz)
-		if err != nil {
-			return nil, err
+		for _, updatedResource := range updatedResources {
```
snehagunta, line 202: Nice.

snehagunta:
One small name change for ResourceReporterRepository, otherwise I think we have everything we need!

internal/biz/resources/resources.go
tylercreller: SGTM - I'll get that updated

tylercreller: updated!

PR #369, opened 2025-03-03
Add v1beta2 protos and wire in service

snehagunta:
## Describe your changes

- ...

Fixes #

## Checklist


internal/service/common/common.go
```diff
@@ -43,6 +44,25 @@ func ResourceFromPb(resourceType, reporterId string, resourceData model.JsonObje
...
+		ResourceType: resourceType,
+		WorkspaceId:  workspaceId,
+		Reporter: model.ResourceReporter{
+			Reporter: model.Reporter{
+				ReporterId:      reporterId,
+				ReporterType:    reporter.ReporterType,
+				ReporterVersion: reporter.ReporterVersion,
+			},
+			LocalResourceId: reporter.LocalResourceId,
+		},
```
tylercreller, line 60: We may need to flatten the Reporter data in the mapper here. In the multi-reporter PR I flattened ReporterResourceID (LocalResourceID) and ReporterID (ReporterType). See the `ResourceFromPbv1beta1` function. Those changes do not look reflected here.

data/schema/common_resource_data.json
tylercreller, line 1: Empty file

api/kessel/inventory/v1beta2/reporter_data.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package kessel.inventory.v1beta2;
+
+option go_package = "github.com/project-kessel/inventory-api/api/kessel/inventory/v1beta2";
+option java_multiple_files = true;
+option java_package = "org.project_kessel.api.inventory.v1beta2";
+
+message ReporterData{
+  string reporter_type = 1;
```
tylercreller, line 10: Should we take this opportunity to rename Reporter Type to Reporter ID?

Today we have...
```
Reporter Type = OCM, ACM, etc.
Reporter ID = "Client data" i.e. Postman, cURL, etc.
```

This seems misleading as an ID should be a unique identifier, like the Service Providers name. I think like we should do the following...
```
Reporter ID = OCM, ACM, etc.
Reporter Type or Reporter Client = "Client Data" i.e. Postman, cURL, etc.
```

internal/service/resources/resources.go
```diff
@@ -0,0 +1,63 @@
...
+	if h, err := toResource(r, identity); err == nil {
+		if resp, err := c.Ctl.Create(ctx, h); err == nil {
+			return fromResource(resp), nil
+
+		} else {
+			return nil, err
+		}
+	} else {
+		return nil, err
+	}
```
tylercreller, line 40: nit: this largely comes down to preference but go teams I have been on have preferred not to use inline error statements to make things more readable, that would look something like...

```
h, err := toResource(r, identity)
if err != nil {
	return nil, err
}

resp, err := c.Ctl.Create(ctx, h)
if err != nil {
	return nil, err
}

return fromResource(resp), nil
```

data/schema/common_resource_data.json
snehagunta: It is empty, yes.But @Adam0Brien is going to add some json schema here, in his PR. It may get merged into this branch in a day or two.

api/kessel/inventory/v1beta2/reporter_data.proto
snehagunta: I have been meaning to get some clarification about this from Alec and Jaime anyway, as to what each of these fields mean and that can inform our naming. I will do that first and see if it aligns with your suggestions.

internal/service/resources/resources.go
snehagunta: I like this as well, compared to a 100 nested if blocks. Update made.

api/kessel/inventory/v1beta2/resource.proto
```diff
@@ -0,0 +1,18 @@
...
+option go_package = "github.com/project-kessel/inventory-api/api/kessel/inventory/v1beta2";
+option java_multiple_files = true;
+option java_package = "org.project_kessel.api.inventory.v1beta2";
+
+message Resource {
+
+ string resource_type = 1 [(buf.validate.field).string = {min_len: 1}];
+ ReporterData reporter_data = 2 [(buf.validate.field).required = true];
+ google.protobuf.Struct common_resource_data = 3 [(buf.validate.field).required = true, json_name = "commonResourceData"];
+
```
alechenninger, line 17: Curious, what can a client do to create a proto struct, given the JSON schema, in a tool-assisted way?

e.g. I can use JSON schema to generate types (e.g in python or go or whatever) which I can use to create valid objects adhering that schema. What's it like to go from that object, to the gRPC API, where it requires an equivalent proto struct?

It would be interested to see if this is something the client libraries could help handle, though that needn't be a requirement for now.

data/schema/resources/k8s_cluster/reporters/acm/config.yaml
```diff
@@ -0,0 +1,13 @@
+---
+resource_type: k8s_cluster
+reporter_name: acm
+namespace: acm
+# this is the name of the field of the resource that contains this reporter's data
```
jmelis, line 5: please remove all this comment section

internal/middleware/preload_schema.go
```diff
@@ -0,0 +1,123 @@
...
+	"path/filepath"
+	"sync"
+)
+
+var schemaCache sync.Map
+
+func PreloadAllSchemas(resourceDir string) error {
+	// Set default resource directory if not provided
+	if resourceDir == "" {
+		resourceDir = "data/schema/resources"
```
jmelis, line 17: I'm not a fan of hardcoding paths so deep inside the codebase. Can we ensure the default is defined at the config level?

internal/middleware/transformer.go
```diff
@@ -0,0 +1,132 @@
...
+
+type ResourcePayload struct {
+	Resource struct {
+		ResourceType       string                 `json:"resource_type"`
+		ReporterData       ReporterDataPayload    `json:"reporter_data"`
+		CommonResourceData map[string]interface{} `json:"common_resource_data"`
+	} `json:"resource"`
+}
+
+func TransformMiddleware() middleware.Middleware {
```
jmelis, line 35: what happens if we don't include this transformer?

I'd like to keep v1beta1 as is, and make v1beta2 only work with the new payload/endpoints. What are the risks of doing it that way?

data/schema/resources/reporter_data.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "reporter_type": {
+      "type": "string",
+      "enum": ["ACM", "HBI", "OCM", "NOTIFICATIONS"]
```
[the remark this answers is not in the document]
jmelis: Ah, down below you have this: `ValidateResourceReporterCombination`. If I understand the code correctly, you are already doing what I described above, so this `enum` is not needed?

internal/middleware/util.go
```diff
@@ -25,3 +31,135 @@ func GetFromContext[V any, I any](i I) func(context.Context) (*V, error) {
 
 	})
 }
+
+func NormalizeResourceType(resourceType string) string {
```
jmelis, line 35: I might be missing something obvious, but, wouldn't it be "easier" if we just keep the filepath -as-is- as the key for the cache? I've seen a bunch of helpers around this key convention, and I'm not sure I understand if it's needed?

internal/middleware/util.go
```diff
@@ -25,3 +31,135 @@ func GetFromContext[V any, I any](i I) func(context.Context) (*V, error) {
...
+
+	resourceData, hasResourceData := reporterData["resourceData"]
+
+	// Case 1: If no schema exists for the resourceType, treat it as an abstract resource and log a warning
+	if err != nil {
+		if hasResourceData {
+			// If resourceData is provided but no schema is found, return error
+			return fmt.Errorf("no schema found for '%s', but 'resourceData' was provided. Submission is not allowed", resourceType)
+		}
+		log.Warnf("no schema found for %s, treating as an abstract resource", resourceType)
```
jmelis, line 125: Do we need this warning? If there is no schema and no resourceData, everything is OK, no? no need to log anything?

internal/middleware/util.go
```diff
@@ -25,3 +31,135 @@ func GetFromContext[V any, I any](i I) func(context.Context) (*V, error) {
...
+		return nil
+	}
+
+	// Case 2: If resourceData is provided, validate it
+	if hasResourceData {
+		if resourceDataMap, ok := resourceData.(map[string]interface{}); ok {
+			// If resourceData is empty but the schema allows it, pass validation
+			if len(resourceDataMap) == 0 {
+				log.Warnf("Resource data for '%s' is empty, but it's allowed by the schema", resourceType)
```
jmelis, line 134: same here, I would suggest to omit this log

jmelis (changes requested):
I left some comments, but more importantly:

- Please add a description to PR, with the changes introduced etc. Something we can share with notifications and ACM so that they understand what this is.
- Let's leave it in draft until we get approval from notifications and ACM

(I'm pretty sure you already have both things in mind, though :) )

I'm super excited about this PR, amazing work everyone! 🚀

internal/middleware/transformer.go
akoserwal: yes, we can remove the middleware

internal/middleware/util.go
akoserwal: The challenge with the `notifications_integration` is that the resource type submitted by the notification team is "resource_type": "notifications/integration," which is different from other types. During the preload logic, a schema cache with key notifications/integration converted to notifications_integration is created. That's why helper function is used. Otherwise we need to add some if/else block for notifications

internal/middleware/util.go
akoserwal: Convert it to debug log would be suitable here

internal/middleware/preload_schema.go
akoserwal: I agree, we can make it configurable.

data/schema/resources/rhel_host/rhel_host.json
```diff
@@ -0,0 +1,12 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "satellite_id": { "type": "string" },
```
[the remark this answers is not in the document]
akoserwal: I agree, if we can involve HBI folks to guide us on the validation rules. We can make it more strict

internal/middleware/transformer.go
Adam0Brien: Addressed in : https://github.com/project-kessel/inventory-api/pull/369/commits/292b932f786c53d2133dab3eea93a6abc2a9e387

internal/middleware/util.go
Adam0Brien: converted to debug in : https://github.com/project-kessel/inventory-api/pull/369/commits/00f2731ad309f2b085f60ba11c4d4db0abbe28f2

internal/middleware/util.go
Adam0Brien: I've removed the log and changed to a comment instead: https://github.com/project-kessel/inventory-api/pull/369/commits/00f2731ad309f2b085f60ba11c4d4db0abbe28f2

data/schema/resources/reporter_data.json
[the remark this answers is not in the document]
Adam0Brien: Good catch! I've made the changes in https://github.com/project-kessel/inventory-api/pull/369/commits/be9c57953efa42e660641ebad1d60c6cd25e4bbf,  removed this file and updated the validation logic to validate resources using the path of `data/schema/resources/{resourceType}/reporters/{reporterType}/{resourceType}.json` 

This means we are now validating in 3 areas, `ResourceReporterCombination`, `ReporterResourceData` and `CommonResourceData`, then ReporterData itself is validated using buf validate at the proto level

I've tested all four resource endpoints using the data from https://github.com/snehagunta/inventory-api/pull/2 and checked both filesystem loading and [json cache loading](https://github.com/snehagunta/inventory-api/blob/v1beta2/schema_cache.json) options

data/schema/resources/k8s_cluster/reporters/acm/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+# relation_properties:
```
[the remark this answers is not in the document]
Adam0Brien: removed in https://github.com/project-kessel/inventory-api/pull/369/commits/8678167f8b4b2d76a20c45d7ef7599ddfa8cc16f

data/schema/resources/rhel_host/rhel_host.json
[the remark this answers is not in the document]
Adam0Brien: updated host schema in https://github.com/project-kessel/inventory-api/pull/369/commits/c3488ebf5220b3ef1f09433f8fea4cba0e29a564

jmelis:
@akoserwal I reviewed the latest logic for Upsert. I have created a diagram to visualize the flow. The white parts is what you already have, that does not require any changes. But I think you should modify your logic to match the green bits.
image

akoserwal:
> @akoserwal I reviewed the latest logic for Upsert. I have created a diagram to visualize the flow. The white parts is what you already have, that does not require any changes. But I think you should modify your logic to match the green bits. image

Updated the logic

akoserwal (approved):

openapi.yaml
```diff
@@ -392,6 +392,85 @@ paths:
...
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/kessel.inventory.v1beta2.DeleteResourceResponse'
+                default:
+                    description: Default error response
+                    content:
+                        application/json:
+                            schema:
+                                $ref: '#/components/schemas/google.rpc.Status'
+    /api/inventory/v1beta2/resources?localResourceId={reporter_type:
```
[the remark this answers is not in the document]
akoserwal:
``` rpc DeleteResource(DeleteResourceRequest) returns (DeleteResourceResponse) {
    option (google.api.http) = {
      delete : "/api/inventory/v1beta2/resources"
      body : "*"
      additional_bindings {
        get: "/api/inventory/v1beta2/resources?local_resource_id={local_resource_id}&reporter_type={reporter_type}"
      }
    };
  };````

Openapi.yaml is generated from ^ above code but looks like it is not generating correctly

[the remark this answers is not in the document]
Comment (snehagunta) [reply openapi.yaml:552]: addressed in https://github.com/project-kessel/inventory-api/pull/427

File: api/kessel/inventory/v1beta2/resource.proto (hunk 4 above)
Comment (snehagunta) [reply api/kessel/inventory/v1beta2/resource.proto:17]: Added a Jira task for this https://issues.redhat.com/browse/RHCLOUD-39149

File: internal/biz/resources/resources.go [hunk 15]
```diff
@@ -60,6 +61,43 @@ func New(reporterResourceRepository ReporterResourceRepository, inventoryResourc
...
+			}
+		}
+		log.Info("Creating resource: ", m)
+		ret, _, err = uc.reporterResourceRepository.Create(ctx, m)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// mock the created at time for eventing
+		// TODO: remove this when persistence is always enabled
```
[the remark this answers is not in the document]
snehagunta: added Jira task to address this https://issues.redhat.com/browse/RHCLOUD-39150

PR #370, opened 2025-03-03
Zookie POC - Check that resource exists

lennysgarage:
## Describe your changes

- When testing out inventory with notifications in ephemeral came across an issue with trying to curl for any integration with error response: "record not found".
- For example when trying to create an integration, we run a checkforcreate call to see if we are allowed to create and then create. Because nothing exists in inventory, this would fail.
- Similar with trying to view any integration, if they don't already exist a 500 error would be returned.

#### SOLUTION:
For `CheckForView` check if the resource exists in inventory DB. if not return not allowed.
For `CheckForUpdate` and `CheckForCreate` check if resource exists in inventory DB. We don't expect anything to exist for create, so create a resource with what we do have available. returns allowed IF allowed.

Issue with `CheckForView` with this is that if the resource doesn't first exist in inventory's DB we will never check if you have permission. ALWAYS disallowed. 
could possibly do same solution as `CheckForCreate` but we don't have a consistency token to send forcing us to make a minimize latency call.

Fixes #

## Checklist

### DIFF
--- a/internal/biz/resources/resources.go
+++ b/internal/biz/resources/resources.go
@@ -102,7 +102,12 @@ func (uc *Usecase) Create(ctx context.Context, m *model.Resource) (*model.Resour
 func (uc *Usecase) CheckForView(ctx context.Context, permission, namespace string, sub *kessel.SubjectReference, id model.ReporterResourceId) (bool, error) {
 	res, err := uc.repository.FindByReporterResourceId(ctx, id)
 	if err != nil {
-		return false, err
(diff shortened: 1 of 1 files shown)


merlante:
I think we should return not allowed if any of the resources don't exist. The reason is, in the `CheckForCreate` case, we should not be checking for the "to be created resource" but for the workspace it will it will created in, which should exist. So I think this is a data population issue.

PR #392, opened 2025-03-18
RHCLOUD-38738: adds kafka consumer

tonytheleg:
## Describe your changes

This change implements a new Kafka Consumer that will monitor our outbox table topic to
* decode tuple requests from the Kafka event
* create the tuple in SpiceDB via Relations API
* capture the consistency token from the Relations API response
* update all records in the Inventory DB sharing the same `inventory_id` with the latest consistency token (zookie)

Other changes made unrelated to the above:
* Updated sample notification payload so subject and resource ID's are easier to differentiate
* Updated Kind/e2e test updates to give it more time as it was timing out due to retry loop
* Updates debezium connector config to support multiple versions of Debezium (ephemeral runs an old one)
* Adds header support to Debezium to capture operation types (for knowing if tuple request is create, delete, or update)
* Make target updates for debezium testing and related sql files

For RHCLOUD-38738

## Checklist

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -220,7 +220,9 @@ func NewCommand(
 			}()
 
 			if !storageOptions.DisablePersistence {
-				go inventoryConsumer.Consume()
+				go func() {
+					_ = inventoryConsumer.Consume()
+				}()
 			}
 
 			quit := make(chan os.Signal, 1)
--- a/internal/consumer/consumer.go
+++ b/internal/consumer/consumer.go
@@ -37,8 +37,8 @@ type KeyPayload struct {
 
 // MessagePayload stores the event message value captured from the topic as emitted by Debezium
 type MessagePayload struct {
-	MessageSchema  map[string]interface{} `json:"schema"`
-	RelationsTuple interface{}            `json:"payload"`
+	MessageSchema    map[string]interface{} `json:"schema"`
+	RelationsRequest interface{}            `json:"payload"`
 }
 
 // New instantiates a new InventoryConsumer
@@ -95,6 +95,7 @@ func (i *InventoryConsumer) Consume() error {
 			case *kafka.Message:
 				// capture the operation from the event headers
 				var operation string
+				var resp interface{}
 				for _, v := range e.Headers {
 					if v.Key == "operation" {
 						operation = string(v.Value)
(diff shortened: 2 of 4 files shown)


cmd/serve/serve.go
```diff
@@ -194,6 +213,8 @@ func NewCommand(
 				srvErrs <- server.Run(ctx)
 			}()
 
+			go consumer.Consume()
```
tonytheleg, line 216: this needs improving either way...possibly leverage existing srvErrs? or just check error period

internal/consumer/consumer.go
```diff
@@ -0,0 +1,186 @@
...
+					continue
+				}
+
+				if !i.PersistenceDisabled {
+					err = i.UpdateConsistencyToken(e.Key, resp)
+					if err != nil {
+						i.Logger.Infof("failed to update consistency token: %v", err)
+						continue
+					}
+				}
```
tylercreller, line 105: IMO we should completely avoid running the consumer altogether if persistence is disabled. While it's super unlikely we don't want the consumer to process messages and potentially just throw the consistency token away if persistence is toggled on and off for some reason. I'm thinking if we disable the consumer completely when persistence is disabled we don't have to worry about any potential side effects. I think this is especially pertinent since outbox messages won't be getting published when persistence is disabled anyway.

For now my thought would be to leave the existing relations replication in place for when persistence is disabled since we do not need a fully consistent solution if persistence is off. Thoughts? I know we talked about this briefly yesterday and this implementation is what we came to haha, so apologies for changing things up.

tonytheleg: > I think this is especially pertinent since outbox messages won't be getting published when persistence is disabled anyway.

😂 excellent point, this all just becomes wasted resources and cycles when persistence is disabled. I'm on board with, ill pull out the logic and struct value since we won't need

internal/consumer/consumer.go
tylercreller, line 1: Just a heads up, we also had a late add onto the design doc around the guarantees we need from the consumer. It's under the "Inventory Consumer Guarantees" section. More of an FYI rather than something specific I noticed.

cmd/serve/serve.go
tonytheleg: Updated this a little bit, ive created a new channel of errors within the consumer similar to how eventing is setup so that if its a consumer error we can explicitly show it is in 'reason'. Also updated this to use a go func for the purpose of at least ignoring the error since errors are being handled elsewhere

internal/consumer/consumer.go
tonytheleg: Added logic to not instantiate the consumer at all if persistence is disabled. Save some memory and execution

internal/consumer/consumer.go
tonytheleg: Add changes to ensure idempotency is covered as mentioned. All Tuple operations should be repeatable.

tonytheleg:
already merged in prior to this branch. closing

PR #405, opened 2025-03-25
RHCLOUD-38543: Consumer change post-zookie updates

tylercreller:
- Updated use of resource update to supply correct params & return
- Updated consumer for new params
- Updated tests for namespace addition


tonytheleg (approved):
LGTM

PR #412, opened 2025-03-26
RHCLOUD-38920 - Check fallback to minimize latency & only record ktn if resource exists

lennysgarage:
## Describe your changes

- Check will fallback to using minimize latency if no resource can be found.
- - Will already use minimize latency in the case that a resource exists but has no ktn (if that's possible?)
- CheckForUpdate will only update the consistency token if the resource already exists, otherwise no side effect.

Fixes # https://issues.redhat.com/browse/RHCLOUD-38920

## Checklist

### DIFF
--- a/internal/biz/resources/resources.go
+++ b/internal/biz/resources/resources.go
@@ -214,11 +214,10 @@ func updateExistingReporterResource(ctx context.Context, m *model.Resource, exis
 func (uc *Usecase) Check(ctx context.Context, permission, namespace string, sub *kessel.SubjectReference, id model.ReporterResourceId) (bool, error) {
 	res, err := uc.reporterResourceRepository.FindByReporterResourceId(ctx, id)
 	if err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
-			// resource doesn't exist.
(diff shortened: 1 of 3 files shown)


internal/biz/resources/resources.go
```diff
@@ -214,11 +214,10 @@ func updateExistingReporterResource(ctx context.Context, m *model.Resource, exis
 func (uc *Usecase) Check(ctx context.Context, permission, namespace string, sub *kessel.SubjectReference, id model.ReporterResourceId) (bool, error) {
 	res, err := uc.reporterResourceRepository.FindByReporterResourceId(ctx, id)
 	if err != nil {
-		if errors.Is(err, gorm.ErrRecordNotFound) {
-			// resource doesn't exist.
-			return false, nil
+		if !errors.Is(err, gorm.ErrRecordNotFound) {
```
merlante, line 217: I think it would be good to leave some generous comments in the code as to what we're doing here, since the intent is not obvious and the changes should probably be reversed in the future when the data is consistent. It would also help to capture the intent for anyone who ends up refactoring this code in the short term.

PR #422, opened 2025-03-28
RHCLOUD-39081: deploy setup updates

tonytheleg:
## Describe your changes

Major overhaul of local development setup configs/manifests. The new structure attempts to clean up the current file structure for better organization and update configurations to model multiple "Test Setups" that a kessel maintainer would generally leverage for various testing needs

Changes:
* Merges all services into a single docker compose file vs one for each test setup
* Adds top level services that define each test-setup (models? interfaces? madness?) which lists the dependent services for that test setup
* defines separate inventory-api config files for each test setup that is leveraged via variable injection during docker compose standup
* All new make targets to deploy each test setup
* Consolidates all the start/stop inventory scripts into a single script for each (except for Kind script)
* Moves all configs/compose setup files into `development` folder path
* Updates README for changes with instructions on how to deploy each test setup, with reference to the config file used
* Adds new consumer settings and logic to allow for disabling the consumer (for local use) or allow for running the consumer without relations (logs but does not attempt to call relations)

For # [RHCLOUD-39081](https://issues.redhat.com/browse/RHCLOUD-39081)

## Checklist


Adam0Brien:
Do we need to configure v1beta2 endpoints for any of the configs?
```sh
resources:
  schemaPath: "data/schema/resources"
  ```

**Comment by tonytheleg** [comment]
> Do we need to configure v1beta2 endpoints for any of the configs?
> 
> ```shell
> resources:
>   schemaPath: "data/schema/resources"
> ```

honestly, I'm sure ill have a bunch of changes to make once all the v1beta2 stuff merges into this branch haha. We're holding on merging this until we merge your changes into here and see what needs to be adjusted.

File: internal/consumer/consumer.go [hunk 1]
```diff
@@ -4,11 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"go.opentelemetry.io/otel"
 	"os"
 	"os/signal"
 	"syscall"
 
+	"github.com/project-kessel/inventory-api/internal/authz/allow"
+	kessel2 "github.com/project-kessel/inventory-api/internal/authz/kessel"
```
tylercreller, line 12: nitpicking here but I don't love the `kessel` naming for imports in these files haha

thoughts on renaming this to `authz` or `relations`?

or even renaming kessel to something like `v1` and keep this as `kessel`

tylercreller (approved):

internal/consumer/consumer.go
tonytheleg: yeah i mean to revisit this, i dont love it either. it got really hard to find something because we import basically all those terms. I tried changing kessel to v1 so that this could be kessel but got errors, lemme see if i can get around it

tonytheleg: fixed -- dont know why it gave me such a hard time last time i tried changing it haha

tylercreller (approved):

PR #430, opened 2025-04-01
Postgres Listen PoC (no notify, only Create)

tylercreller:
- Implemented Listen/Notify PoC
  - Listens to a single pg channel `consumer-notifications` on app startup
  - Subscribers to `ListenManager` determine what notifications they care about based on a generated transaction ID (`txId`)
  - Request handlers create subscriptions to receive notifications and block the request context

### Components
**Driver:** Manages postgres connection and initiates `LISTEN`

**ListenManager:** Initiates `waitForNotification` via `Driver`. Manages subscriptions for all requests handlers. Receives and distributes notifications to subscriptions

**Subscriptions:** Created by request handlers via `ListenManager`. Have their own golang channels to receive notifications.

### TODO:
* Handle update & upsert (v1beta2) requests
* Handle automatic reconnects in case postgres goes down
* Remove debugging logs
* Do not crash when using sqlite

### DIFF
--- a/internal/pubsub/listenmanager.go
+++ b/internal/pubsub/listenmanager.go
@@ -1,0 +1,144 @@
+package pubsub
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/go-kratos/kratos/v2/log"
+)
+
+type ListenManager interface {
+	Subscribe(txId string) Subscription
+	Run(ctx context.Context) error
+}
+
+type Subscription interface {
+	NotificationC() <-chan []byte
+	Unsubscribe(ctx context.Context)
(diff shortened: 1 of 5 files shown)


internal/pubsub/listenmanager.go
```diff
@@ -0,0 +1,144 @@
...
+}
+
+type channelChange struct {
+	channel   string
+	close     func()
+	operation string
+}
+
+// Listen returns a Subscription.
+// TODO change to Subscribe
```
lennysgarage, line 71: Is this TODO still relevant?

tylercreller: Nope! I'll remove it during cleanup

PR #432, opened 2025-04-01
RHCLOUD-38738: Add consumer retry

tonytheleg:
## Describe your changes

Bug Fix: When the consumer reads a message from Kafka, but fails to process (in this case a failed Relations call), the failed message is never reprocessed despite not being committed. This is because Kafka, once it consumes the message, updates its position and only reads new messages sent. In order to handle retrying of a message, you much implement retry logic on the call itself, or re-start the consumer so that it uses the last offset committed which would replay the failed message. Also important, any messages committed after the failed message would cause that message to be lost forever so the consumer must fail if it cannot process the message.

Changes to address bug:
* adds operation and consumer retry processing
  * adds new retry options and config to set retry values (allow for changing via deployment if we need to increase time)
  * runs all API requests within new Retry function to handle retries and backoff
  * ensures repetitive failures forces the consumer to restart without killing the service
  * eventually multiple consumer restarts leads to killing the service

Also, Improvements:
* refactors message parsing into their own functions so they are easier to test outside of the API request calls
* adds test for message parsing
* adds new config and make target for running a split setup with relations for testing locally with docker (doc updates too)
* removes toolchain directive to enforce go 1.22.9 to remain FIPS compliant

Fixes #

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,6 +161,10 @@ inventory-up-relations-ready:
 inventory-up-split:
 	./scripts/start-inventory.sh split-setup 8000 9000
 
+.PHONY: inventory-up-split-relations-ready
+inventory-up-split-relations-ready:
+	./scripts/start-inventory.sh split-setup-relations-ready 8081 9081
+
 .PHONY: inventory-up-sso
 inventory-up-sso:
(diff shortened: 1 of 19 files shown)


tylercreller:
LGTM

tylercreller (approved):

PR #437, opened 2025-04-02
RHCLOUD-38740- Add ability to toggle read-after-write requests

lennysgarage:
## Describe your changes
v1beta2 changes:
- Updates `ReportResourceRequest` with new bool `wait_for_sync` to enable per resource read after write capability

v1beta1 changes:
- `RhelHost` with new bool `wait_for_sync` to enable per resource read after write capability 
- `NotificationsIntegration` with new bool `wait_for_sync` to enable per resource read after write capability
- `K8sPolicy` with new bool `wait_for_sync` to enable per resource read after write capability
- `K8sCluster` with new bool `wait_for_sync` to enable per resource read after write capability

#
- Adds 2 new server config options to the consumer, `read-after-write-enabled` & `read-after-write-allowlist`
- - `read-after-write-enabled` if false will disable read after write flow for all requests
- - if `read-after-write-enabled` is true & SP is in `read-after-write-allowlist`, create, update & upserts requests will go through the read after write flow. 
- Some updates so the consumer config is available to the usecase struct.

Fixes # 
https://issues.redhat.com/browse/RHCLOUD-38740
https://issues.redhat.com/browse/RHCLOUD-38741 

## Checklist

## Summary by Sourcery

Add server-side configuration to control read-after-write consistency workflow for service providers

New Features:
- Introduce two new configuration options for read-after-write consistency: `read-after-write-enabled` and `read-after-write-allowlist`

Enhancements:
- Extend the consumer and usecase configurations to support selective read-after-write functionality
- Add a mechanism to check if a service provider is in the read-after-write allowlist

## Summary by Sourcery

Add server-side configuration to control read-after-write consistency workflow for service providers

New Features:
- Introduce two new configuration options for read-after-write consistency: `read-after-write-enabled` and `read-after-write-allowlist`

Enhancements:
- Extend the consumer and usecase configurations to support selective read-after-write functionality
- Add a mechanism to check if a service provider is in the read-after-write allowlist

## Summary by Sourcery

Add server-side configuration to control read-after-write consistency workflow for service providers

New Features:
- Introduce two new configuration options for read-after-write consistency: `read-after-write-enabled` and `read-after-write-allowlist`

Enhancements:
- Extend the consumer and usecase configurations to support selective read-after-write functionality
- Add a mechanism to check if a service provider is in the read-after-write allowlist


tonytheleg (commented):
LGTM!

internal/biz/resources/resources.go
```diff
@@ -118,12 +122,19 @@ func (uc *Usecase) Upsert(ctx context.Context, m *model.Resource) (*model.Resour
...
-			timeoutCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
-			defer cancel()
 
-			err = subscription.BlockForNotification(timeoutCtx)
-			if err != nil {
-				return nil, err
+			// read after write functionality is enabled globally.
+			// And executed if request specifies or
+			// request came from service provider in allowlist
+			if uc.ReadAfterWriteEnabled && (requestReadAfterWrite || isSPInAllowlist(m, uc.ReadAfterWriteAllowlist)) {
```
tylercreller, line 129: We will need to also add this condition to the other block that nil checks the ListenManager above. We don't want to subscribe if it's not requested by the client.

Perhaps we could throw it into a bool for share-ability.

```
readAfterWriteEnabled := uc.ListenManager != nil && uc.ReadAfterWriteEnabled && (requestReadAfterWrite || isSPInAllowlist(m, uc.ReadAfterWriteAllowlist))
```

api/kessel/inventory/v1beta2/resource_service.proto
tylercreller, line 1: We should update v1beta1 protos as well, until I can loop back with Sneha.

deploy/kessel-inventory-ephem-w-debezium.yaml
```diff
@@ -20,6 +20,8 @@ objects:
         consumer:
           bootstrap-servers: inventory-kafka-kafka-bootstrap:9092
           topic: outbox.event.kessel.tuples
+          read-after-write-enabled: true # false == off for all service providers
+          read-after-write-allowlist: [] # specify "*" to require all requests
```
tylercreller, line 24: I don't know if `consumer` is the right spot for this config, since it's not actually being leveraged by the consumer. Perhaps we should add a `consistency` section. Thoughts?

api/kessel/inventory/v1beta2/resource_service.proto
```diff
@@ -12,6 +12,7 @@ option java_package = "org.project_kessel.api.inventory.v1beta2";
 
 message ReportResourceRequest {
   Resource resource = 1 [  json_name = "resource" ];
+  bool read_after_write_enabled = 2;
```
tylercreller, line 15: I'm wondering if we should go for a more simple name for this attribute. Something that folks could immediately recognize and understand like `ensure_consistency`, `wait_for_sync`, `synchronous`, or anything else you could come up within the same vein. I'm worried the current name may come back to bite us. IMO it's important to indicate that using this feature may introduce extra latency because we are making the async processes into sync processes before the client will receive a response.

tonytheleg: +1, if read-after-write is not the default, maybe calling it `async` and default to true since that would be our standard process flow and you have to explicitly turn that off for read after write?

deploy/kessel-inventory-ephem-w-debezium.yaml
tonytheleg: the consumer would need to know about whether r-a-w is enabled or not to send the proper notification but i could see the argument that its not just the consumer that cares about that setting.

internal/biz/resources/resources.go
lennysgarage: Missed the subscription part thanks! That bool's a great idea

deploy/kessel-inventory-ephem-w-debezium.yaml
lennysgarage: I'm not against adding a `consistency` section, and it may not be leveraged directly by the consumer, but I was thinking that if the `consumer` was removed we wouldn't have a reason to keep the `read-after-write` features themselves and so couple them together.

api/kessel/inventory/v1beta2/resource_service.proto
lennysgarage: I liked the idea of doing `async` but we can't seem to set a default value for proto3 in the rpc itself, we could make it such that if nothing is specified its async and only when setting a value or forcing false like `  bool async = 2 [(buf.validate.field).bool.const = false];` it's synchronous but its starting to feel more complicated to reason about. The nice thing with proto3 is every field is already optional so if SPs dont set anything then we can go with a default route. I think something like `bool wait_for_sync = 2;` makes sense here. As a default it will be off if nothing is set and only when set to true does the request actually wait to sync.

lennysgarage:
@sourcery-ai review

deploy/kessel-inventory-ephem-w-debezium.yaml
tylercreller: I think my hang-up is that read-after-write is more of an API feature than a consumer feature. We could support read-after-write without the consumer we'd just need to devise a different solution for the pub sub (kafka or something)

Maybe it makes sense in the new `resources` section?  Like you mentioned in slack, this feature really is only applicable to resources.

tylercreller (approved):
LGTM

PR #451, opened 2025-04-07
Remove authz package

snehagunta:
## Describe your changes

- Remove authz package and put KesselCheckService, KesselLookupService and KesselResourceService in a single package

Fixes https://issues.redhat.com/browse/RHCLOUD-39315

## Checklist

## Summary by Sourcery

Consolidate authorization-related services into the resources package by removing the separate authz package

Enhancements:
- Refactor service organization by moving KesselCheckService, KesselLookupService, and related services from the authz package to the resources package

Chores:
- Remove authz package and update import references
- Rename some service initialization methods to be more descriptive


alechenninger:
Overall LGTM.

I do agree with the robot – its weird that the naming convention is different between the different services' constructors. Check includes the version, but not the name (`NewV1beta2`. Others include the name, but not the version (`NewKesselLookupService`). I get that the others only have one version, but I would just always include the version anyway, since the whole point is that we allow ourselves to add versions over time. I would always include the version in these names. Whether or not we include the name of the thing, also, despite it being implied by the import, is up to you. But I would at least include the version regardless and make them consistent otherwise.

Beyond that there is more to do to improve the API of course but this is one step in the right direction IMO.

snehagunta:
@alechenninger I updated the constructor names for v1beta1 and v1beta2. 

This also got me thinking about versioning for health APIs. They are currently versioned but I am seeing suggestions online saying that they are better unversioned. Thoughts?

alechenninger:
The probes are just matching kubernetes so yeah it's not really an API we own to version. They're not a part of "Kessel's" API - it's part of the operational protocol which is its own thing.

I wouldn't bother versioning those though it's not really a big deal either way.

PR #470, opened 2025-04-10
Internal eventual consistency

lennysgarage:
## Describe your changes

- ...

Fixes #

## Checklist


tylercreller (approved):
LGTM

api/kessel/inventory/v1beta2/report_resource_request.proto
```diff
@@ -11,4 +11,6 @@ option java_package = "org.project_kessel.api.inventory.v1beta2";
 
 message ReportResourceRequest {
   Resource resource = 1 [(buf.validate.field).required = true];
+  // Will wait for synchronous processes to complete before returning
+  bool wait_for_sync = 2;
```
alechenninger, line 15: Let's make this more about the contract (the caller can read their writes after the call returns) and less about the implementation (waiting for some external synchronization process). Can you think of better names for the field that resemble that?

A WriteConsistency or WriteConcern enum might make sense, where possible values can be like "minimize_latency" (no promises) or "read_your_write" (ensure reads consistent with this write) or something like that.

Other ideas?

PR #471, opened 2025-04-10
RHCLOUD-39386: v1beta2 support

tylercreller:
- Adds ReporterType mapping to namespace for v1beta2 on outbox events
- Adds test for new ReporterType / Outbox mapping 
- Isolates consistency txids to only generate on readAfterWrite enabled requests

## Summary by Sourcery

Implement v1beta2 support for outbox events and improve transaction ID generation for read-after-write scenarios

New Features:
- Add support for v1beta2 namespace mapping using ReporterType

Enhancements:
- Optimize transaction ID generation to only create when read-after-write is enabled
- Modify outbox event generation to support v1beta2 namespace override

Tests:
- Add test cases for v1beta2 outbox event generation with ReporterType namespace mapping

### DIFF
--- a/internal/biz/resources/resources.go
+++ b/internal/biz/resources/resources.go
@@ -82,14 +82,7 @@ func (uc *Usecase) Upsert(ctx context.Context, m *model.Resource, wait_for_sync 
 	log.Info("upserting resource: ", m)
 	ret := m // Default to returning the input model in case persistence is disabled
 	var subscription pubsub.Subscription
-
-	// Generate txid for data layer
-	// TODO: Replace this when inventory api has proper api-level transaction ids
-	txid, err := uuid.NewV7()
-	if err != nil {
-		return nil, err
-	}
-	txidStr := txid.String()
+	var txidStr string
 
 	if !uc.DisablePersistence {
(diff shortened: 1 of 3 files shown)


internal/biz/resources/resources.go
```diff
@@ -82,14 +82,7 @@ func (uc *Usecase) Upsert(ctx context.Context, m *model.Resource, wait_for_sync
...
 	var subscription pubsub.Subscription
-
-	// Generate txid for data layer
-	// TODO: Replace this when inventory api has proper api-level transaction ids
-	txid, err := uuid.NewV7()
-	if err != nil {
-		return nil, err
-	}
-	txidStr := txid.String()
+	var txidStr string
```
lennysgarage, line 85: Just to ensure understanding, before we always generated a txID and now we are only generating one if readAfterWrite is enabled? For instance on line 110, it would have a blank transaction id because we are updating an existing resource

lennysgarage: Okay yup, reread your summary. Makes sense!

lennysgarage (approved):
LGTM

PR #514, opened 2025-04-25
Set debezium poll.interval.ms to 250ms

tylercreller:
- Sets debezium poll interval to 250ms
  - defines a happy medium between performance and resource utilization until we can get some practical use data

image
Benchmarks:

- 1 pod
- 8 min runs

## Summary by Sourcery

Deployment:
- Update the `poll.interval.ms` setting to 250 across Debezium connector configurations for various environments (production, ephemeral, kind, development).

### DIFF
--- a/deploy/debezium/debezium-connector.yaml
+++ b/deploy/debezium/debezium-connector.yaml
@@ -27,6 +27,7 @@ objects:
       heartbeat.interval.ms: ${DEBEZIUM_HEARTBEAT_INTERVAL_MS}
       heartbeat.action.query: ${DEBEZIUM_ACTION_QUERY}
       topic.heartbeat.prefix: ${TOPIC_HEARTBEAT_PREFIX}
+      poll.interval.ms: ${DEBEZIUM_POLL_INTERVAL_MS}
 parameters:
 - name: KAFKA_CONNECT_INSTANCE
   value: platform-kafka-connect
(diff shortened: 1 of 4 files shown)


deploy/debezium/debezium-connector.yaml
```diff
@@ -27,6 +27,7 @@ objects:
       heartbeat.interval.ms: ${DEBEZIUM_HEARTBEAT_INTERVAL_MS}
       heartbeat.action.query: ${DEBEZIUM_ACTION_QUERY}
       topic.heartbeat.prefix: ${TOPIC_HEARTBEAT_PREFIX}
+      poll.interval.ms: 250
```
lennysgarage, line 30: I feel like we should define this as a parameter like `${DEBEZIUM_HEARTBEAT_INTERVAL_MS}`. Possibly `DEBEZIUM_POLL_INTERVAL_MS`?

deploy/kessel-inventory-ephem.yaml
```diff
@@ -257,6 +257,7 @@ objects:
         heartbeat.interval.ms: ${DEBEZIUM_HEARTBEAT_INTERVAL_MS}
         heartbeat.action.query: ${DEBEZIUM_ACTION_QUERY}
         topic.heartbeat.prefix: ${TOPIC_HEARTBEAT_PREFIX}
+        poll.interval.ms: 250
```
lennysgarage, line 260: Same here, we could use a parameter to set the poll rate.

deploy/debezium/debezium-connector.yaml
tylercreller: Good call, lemme do that

tonytheleg (approved):
LGTM

lennysgarage (approved):
LGTM

PR #523, opened 2025-04-28
RHCLOUD-39583: Update Optional fields in protos everywhere to have optional keyword

Adam0Brien:
## Describe your changes

- Add optional keyword to optional proto fields

Fixes: https://issues.redhat.com/browse/RHCLOUD-39538

## Checklist

## Summary by Sourcery

Enhancements:
- Update Protobuf definitions to explicitly mark the `relation` field as optional.


api/kessel/inventory/v1beta2/check_for_update_response.proto
```diff
@@ -9,5 +9,5 @@ option java_multiple_files = true;
 option java_package = "org.project_kessel.api.inventory.v1beta2";
 
 message CheckForUpdateResponse {
-  Allowed allowed = 1;
+  optional Allowed allowed = 1;
```
alechenninger, line 12: This would be present always, so doesn't make sense to have optional presence

The validation might not make sense, because it's a response? Is it typical to add validation on response objects? Maybe?

alechenninger: Looking through some other protos, I don't think adding validation on responses is typical. So, I think this (and other response protos) will not have explicit presence, because they are not actually optional semantically. We just don't add the validate expressions because it doesn't really have value since these are not inputs. They are outputs.

snehagunta: Updated.

akoserwal (approved):

PR #545, opened 2025-05-02
RHCLOUD-39799: Derive namespace from resource/request for outbox event tuples

tylercreller:
- Support deriving tuple namespace from resources rather than request handlers when possible
- Will still fallback to the handler namespace if the resource does not contain complete data

## Summary by Sourcery

Improve namespace derivation for outbox events by prioritizing resource-level namespace information

Bug Fixes:
- Remove hardcoded namespace workaround for 'authz' to 'notifications' mapping

Enhancements:
- Modify namespace derivation logic to first check resource-level reporter type before falling back to handler namespace

Tests:
- Update test cases to use resource reporter type for namespace validation


tonytheleg (approved):
LGTM

PR #562, opened 2025-05-06
Update consistency param

tylercreller:
- Updates `WaitForSync` v1beta2 param to `WriteVisibility`
- Bumps dockerfile go versions to `1.23.6`

## Summary by Sourcery

Update the consistency parameter in the inventory API from `WaitForSync` to a more granular `WriteVisibility` enum, providing more explicit control over write operations and their consistency semantics.

New Features:
- Introduced a new `WriteVisibility` enum with three modes: UNSPECIFIED, MINIMIZE_LATENCY, and IMMEDIATE

Enhancements:
- Replaced boolean `WaitForSync` with a more expressive `WriteVisibility` enum
- Updated read-after-write logic to use the new write visibility parameter

Build:
- Bumped Dockerfile Go version to 1.23.6

Documentation:
- Updated documentation for read-after-write enablement to reflect new write visibility mechanism


snehagunta (approved):
Small nit. Otherwise LGTM.

tylercreller:
> You may need to run `make api_breaking` if you haven't already?

Proto updated the parameter name, needed to apply those to the tests and logic

Dockerfile-e2e
```diff
@@ -5,8 +5,8 @@ USER root
 RUN microdnf install -y tar gzip make which gcc gcc-c++ cyrus-sasl-lib findutils git
 
 # install platform specific go version
-RUN curl -O -J  https://dl.google.com/go/go1.22.5.linux-${TARGETARCH}.tar.gz
-RUN tar -C /usr/local -xzf go1.22.5.linux-${TARGETARCH}.tar.gz
+RUN curl -O -J  https://dl.google.com/go/go1.23.6.linux-${TARGETARCH}.tar.gz
```
tonytheleg, line 8: Didnt realize e2e was still using upstream go. Possibly we should switch these out to go-toolset for simplicity? I dont think we have to force the FIPS build or anything since it won't run on  FIPS system but we would never have to change the GO version again which would save time in future? Not a big deal for now, but something we should consider

tonytheleg (approved):
No strong opinions on the WaitForSync change -- just made a comment on still using upstream Go in e2e dockerfile

Dockerfile-e2e
tylercreller: Yeah good call out. I think that's the play, at least to keep things consistent and accurate in comparison to our live envs. I'll get a card in the backlog.

lennysgarage: Oh yeah I must've missed this when I was updating relations because that was exactly what I did there, https://github.com/project-kessel/relations-api/blob/main/api/Dockerfile makes things a lot simpler as well.

lennysgarage (approved):
LGTM

PR #568, opened 2025-05-07
RHCLOUD-39802: Return 200 for listen/notify timeouts

tylercreller:
- If context is cancelled via timeout, return 200 for requests

## Summary by Sourcery

Modify listen/notify timeout handling to return a 200 status code when the context is cancelled

Bug Fixes:
- Handle context cancellation by returning a successful response instead of an error in resource operations

Enhancements:
- Introduce a new error type ErrWaitContextCancelled for clearer context timeout handling


lennysgarage (approved):
LGTM

PR #571, opened 2025-05-07
RHCLOUD-39804 - Add an indefinite retry mechanism to the consumer

lennysgarage:
## Describe your changes

- Add a mechanism to indefinitely retry the consumer itself 
- Add a mechanism to indefinitely retry processing a failing message
- Add a configurable maximum amount of backoff time between retries
```
retry-options:
    consumer-max-retries: -1
    operation-max-retries: -1
    max-backoff-seconds: 30
```
- Updates ephemeral deployment with indefinite retries & backoff setting
- Adds some metrics counters to the Retry().

Fixes # https://issues.redhat.com/browse/RHCLOUD-39804

## Checklist

## Summary by Sourcery

Implement an indefinite retry mechanism for the Kafka consumer to improve resilience and error handling

New Features:
- Add support for indefinite retries for both consumer and message processing by using -1 as a retry count

Enhancements:
- Implement a maximum backoff time of 30 seconds to prevent excessive wait times
- Add metrics tracking for retry failures

Deployment:
- Update ephemeral deployment configuration to enable indefinite retries

## Summary by Sourcery

Implement an indefinite retry mechanism for the Kafka consumer to improve system resilience and error handling

New Features:
- Add support for indefinite retries for both consumer and message processing by using -1 as a retry count

Bug Fixes:
- Prevent message dropping by restarting consumer on processing failures

Enhancements:
- Implement a maximum backoff time of 30 seconds to prevent excessive wait times
- Add metrics tracking for retry failures

Deployment:
- Update ephemeral deployment configuration to enable indefinite retries

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -288,6 +288,7 @@ func NewCommand(
 			shutdown := shutdown(db, server, eventingManager, &inventoryConsumer, log.NewHelper(logger))
 
 			if !storageOptions.DisablePersistence && consumerOptions.Enabled {
+				const maxBackoff = 30 * time.Second
 				go func() {
 					retries := 0
 					for consumerOptions.RetryOptions.ConsumerMaxRetries == -1 || retries < consumerOptions.RetryOptions.ConsumerMaxRetries {
@@ -303,7 +304,7 @@ func NewCommand(
 							inventoryConsumer.Logger.Errorf("consumer unable to process current message -- restarting consumer")
 							retries++
 							if consumerOptions.RetryOptions.ConsumerMaxRetries == -1 || retries < consumerOptions.RetryOptions.ConsumerMaxRetries {
-								backoff := time.Duration(inventoryConsumer.RetryOptions.BackoffFactor*retries*300) * time.Millisecond
+								backoff := min(time.Duration(inventoryConsumer.RetryOptions.BackoffFactor*retries*300)*time.Millisecond, maxBackoff)
 								inventoryConsumer.Logger.Errorf("retrying in %v", backoff)
 								time.Sleep(backoff)
 							}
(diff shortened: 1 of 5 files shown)


internal/consumer/consumer.go
```diff
@@ -183,6 +183,7 @@ func (i *InventoryConsumer) Consume() error {
 
 				resp, err = i.ProcessMessage(headers, relationsEnabled, e)
 				if err != nil {
+					Incr(i.MetricsCollector.msgProcessFailures, "ProcessMessage", err)
```
tonytheleg, line 186: Double check me here, but i think i left this off because there were a few places where ProcessMessage could fail inside that function so instead of doing it at the top level, i only did it inside the fuction so that i could point out the exact function inside process message that failed. I think this would increment it a second time on each failure?

lennysgarage: Yup, you're spot on it would increment twice as each, as each create, update, and delete already have a `Incr()` call.

cmd/serve/serve.go
```diff
@@ -288,9 +288,10 @@ func NewCommand(
 			shutdown := shutdown(db, server, eventingManager, &inventoryConsumer, log.NewHelper(logger))
 
 			if !storageOptions.DisablePersistence && consumerOptions.Enabled {
+				const maxBackoff = 30 * time.Second
```
tylercreller, line 291: I see we are leveraging this in a couple places, this might actually be a good candidate to be configurable via the consumerOptions. Thoughts?

internal/consumer/consumer.go
lennysgarage: Gonna go ahead and fix that, thanks!

cmd/serve/serve.go
tonytheleg: also not a bad idea

lennysgarage: I think it may make sense tho as a configurable option so we could tune down the wait without having to rebuild if we needed to debug issues quickly with the consumer

tonytheleg (approved):
LGTM!

tylercreller (approved):
LGTM

PR #579, opened 2025-05-12
RHCLOUD-39803: Notification circuit breaker for consumer health

tylercreller:
## Summary
- Adds wait for notification circuit breaker to v1beta2 `Upsert`
- Circuit breaker behavior
  - If we reach the target number of consecutive failed waits (3)
     - Trip the circuit breaker to open (no waiting for notifs)
     - After the timeout period, the circuit breaker will be half-open and one request will be allowed through
     - If it succeeds, the circuit breaker will fully close (wait for notifs)
     - If it fails, it will go back to open

## Pros
- Easy to implement and maintain
- Consumer implementation agnostic
    - If we ever move the consumer out of the inventory API container, there won’t be any work needed here.

## Cons
- Reactive. Requires some time to realize the consumer is down, as well as recover.
- Circuit breaker is per-pod so depending on routing and load, failures could take longer to be realized and short-circuit.

### DIFF
--- a/cmd/serve/serve.go
+++ b/cmd/serve/serve.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/go-kratos/kratos/v2/log"
+	"github.com/sony/gobreaker"
 	"github.com/spf13/cobra"
 	"gorm.io/gorm"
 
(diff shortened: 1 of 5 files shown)


tonytheleg (approved):
LGTM

lennysgarage (approved):
LGTM

tonytheleg (approved):
LGTM

PR #584, opened 2025-05-14
RHCLOUD-39947: validate StreamedListObject

Adam0Brien:
## Describe your changes

- Warn users to when they submit invalid resource and reporter types to streaming endpoint

This is because of the spicedb regex constraint
```sh
value does not match regex pattern "^([a-z][a-z0-9_]{1,61}[a-z0-9]/)*[a-z][a-z0-9_]{1,62}[a-z0-9]$"
```

Instead of letting the request reach relations and SpiceDB, this PR warns the user before the request reaches relations

Valid example:
```sh
grpcurl -plaintext -d \ '{
  "objectType": {
    "resourceType": "hOsT",
    "reporterType": "hBi"
  },
  "relation": "view",
  "subject": {
    "resource": {
      "resourceType": "principal",
      "resourceId": "sarah",
      "reporter": {
        "type": "rbac"
      }
    }
  }
}' "localhost:9081" kessel.inventory.v1beta2.KesselInventoryService.StreamedListObjects

# This will show up in relations as :  args=resource_type:{namespace:"hbi" name:"host"} relation:"view" subject:{subject:{type:{namespace:"rbac" name:"principal"} id:"sarah"}}
```
Valid example 2:
```sh
grpcurl -plaintext -d \ '{
  "objectType": {
    "resourceType": "host",
    "reporterType": "hbi"
  },
  "relation": "view",
  "subject": {
    "resource": {
      "resourceType": "principal",
      "resourceId": "sarah",
      "reporter": {
        "type": "rbac"
      }
    }
  }
}' "localhost:9081" kessel.inventory.v1beta2.KesselInventoryService.StreamedListObjects
```

Example invalid request (bad reporterType):
```sh
grpcurl -plaintext -d \ '{
  "objectType": {
    "resourceType": "host",
    "reporterType": "hBi!"
  },
  "relation": "view",
  "subject": {
    "resource": {
      "resourceType": "principal",
      "resourceId": "sarah",
      "reporter": {
        "type": "rbac"
      }
    }
  }
}' "localhost:9081" kessel.inventory.v1beta2.KesselInventoryService.StreamedListObjects

ERROR:
  Code: Unknown
  Message: failed to build lookup request: reporter type does not match required pattern: hbi!

```
Example invalid request (bad resourceType):
```sh
 grpcurl -plaintext -d \ '{
  "objectType": {
    "resourceType": "hos!t",
    "reporterType": "hBi"
  },
  "relation": "view",
  "subject": {
    "resource": {
      "resourceType": "principal",
      "resourceId": "sarah",
      "reporter": {
        "type": "rbac"
      }
    }
  }
}' "localhost:9081" kessel.inventory.v1beta2.KesselInventoryService.StreamedListObjects

ERROR:
  Code: Unknown
  Message: failed to build lookup request: resource type does not match required pattern: hos!t

```
Fixes #https://issues.redhat.com/browse/RHCLOUD-39947

## Checklist

## Summary by Sourcery

Validate and normalize resource and reporter types in StreamedListObjects before forwarding lookup requests, returning an error for nil or invalid inputs.

Enhancements:
- Introduce normalizeType helper to lowercase object and reporter types.
- Update toLookupResourceRequest to error on nil requests and apply normalized types when constructing lookup requests.

Tests:
- Add unit tests for normalizeType and IsValidType to verify type normalization and regex pattern matching.


alechenninger:
In the case of reporter type, could we just say that these are case insensitive and lowercase them ourselves when interacting with Relations?

Adam0Brien:
Hey @alechenninger I've added a change to the streming middleware that lowercase's ResourceType and reporterType before the Regex check takes place, This way if we can have "HBI" -> "hbi" and failure cases like "HBI!!!" -> "hbi!!!" still fail as expected

alechenninger:
Thanks @Adam0Brien . WDYT about relaxing the regex, and moving that business logic of case insensitivity out of middleware and into the application (so the method where we process the request input)?

In a sense the regex validation is sort of "lying" that it has to be lowercase, but actually uppercase is accepted (sneakily). This can be problematic if, for example, client-library level validation is used. The client still gets there error and it never makes it to the server, even though the server would accept it and it would be perfectly fine.

Does that makes sense? What do you think?

snehagunta:
@alechenninger I asked @Adam0Brien to move the case insensitivity logic [here](https://github.com/project-kessel/inventory-api/blob/main/internal/authz/kessel/kessel.go), so the anti corruption layer where we transform from Inventory API concepts to spicedb. What do you mean by `(so the method where we process the request input)?`, do you mean where we process the Relations request input or something else?

akoserwal (approved):

PR #588, opened 2025-05-15
minor dashboard fix plus adds dashboards to local compose setup

tonytheleg:
## Describe your changes

Fixes kafka connect dashboard:
* i noticed our kafka connect dashboard was not setting the datasources to use our datasource variable which would not work in prod; fixes it so the same dashboard can be used in stage and prod and datasource can be changed

Updates docker compose to allow for better testing of dashboard changes
* adds makefile target and script to dump the dashboard configmaps to json and pre-load them with grafana when running it in docker compose
* same script also makes sure to update the datasource to match the pre-loaded datasource as well
* updates job name in prometheus config to kessel-inventory-api  to ensure existing dashboards load with label searching
* updates docs on using new make target and how to test dashboard changes

Not all metrics are going to work (specifically anything pointing to MSK) but this does make it a little easier to test dashboard changes for our API with metrics available from the app itself


tylercreller (approved):
LGTM

PR #615, opened 2025-05-22
Fix serialization bug with inventory_id

tylercreller:
- Fixes a bug where retries on a create transaction would not rollback the inventory ID on the resource

### DIFF
--- a/internal/data/resources/resourcerepository.go
+++ b/internal/data/resources/resourcerepository.go
@@ -44,6 +44,10 @@ func copyHistory(m *model.Resource, id uuid.UUID, operationType model.OperationT
 }
 
 func (r *Repo) Create(ctx context.Context, m *model.Resource, namespace string, txid string) (*model.Resource, error) {
+	if m == nil {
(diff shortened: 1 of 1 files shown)


lennysgarage (approved):
LGTM

internal/data/resources/resourcerepository.go
```diff
@@ -47,46 +47,48 @@ func (r *Repo) Create(ctx context.Context, m *model.Resource, namespace string,
 	db := r.DB.Session(&gorm.Session{})
 	var result *model.Resource
 	err := r.handleSerializableTransaction(db, func(tx *gorm.DB) error {
+		// Copy the resource to avoid modifying the original, necessary for serialized transaction retries
+		resource := *m
```
akoserwal, line 51: can `*m` can be a nil?

tylercreller: Practically it will never be nil given the nature of what we need to accomplish with the create function - creating a resource from that model. Hypothetically, it could but it would be a bug. I'll add some nil checks just for completeness.

PR #624, opened 2025-05-28
Fix create log to use returned resource

tylercreller:
- Use the returned resource from the data layer for logging

## Summary by Sourcery

Bug Fixes:
- Update Create usecase to log the ID and type from the returned resource instead of the input model


lennysgarage (approved):
LGTM

PR #685, opened 2025-06-27
V1beta2 report resource

snehagunta:
## Describe your changes

- ...

## Ticket reference (if applicable)
Fixes #

## Checklist

* [ ] Are the agreed upon acceptance criteria fulfilled?

* [ ] Was the 4-eye-principle applied? (async PR review, pairing, ensembling)

* [ ] Do your changes have passing automated tests and sufficient observability?

* [ ] Are the work steps you introduced repeatable by others, either through automation or documentation?
  * [ ] If automation is possible but not done due to other constraints, a ticket to the tech debt sprint is added
  * [ ] An SOP (Standard Operating Procedure) was created

* [ ] The Changes were automatically built, tested, and  - if needed, behind a feature flag - deployed to our production environment. (**Please check this when the new deployment is done and you could verify it.**)

* [ ] Are the agreed upon coding/architectural practices applied?

* [ ] Are security needs fullfilled? (e.g. no internal URL)

* [ ] Is the corresponding Ticket in the right state? (should be on "review" now, put to done when this change made it to production)

* [ ] For changes to the public API / code dependencies: Was the whole team (or a sufficient amount of ppl) able to review?

## Summary by Sourcery

Add v1beta2 reporting resource support by defining new data models, repository implementations, database migrations, transaction utilities, and comprehensive unit tests.

New Features:
- Introduce v1beta2 models: Resource, RepresentationReference (with unique composite index and Postgres covering indexes), CommonRepresentation, ReporterRepresentation, ResourceWithReferences aggregate, and ReporterRepresentationId.
- Implement real and in-memory fake repositories for ResourceWithReferences (including find-by-reporter lookup), CommonRepresentation, and ReporterRepresentation with transactional create operations.
- Integrate v1beta2 models into the GORM auto-migration setup.

Enhancements:
- Add WithTx helper to manage transactions uniformly across v1beta2 repositories.

Tests:
- Add extensive unit tests for real and fake v1beta2 repositories covering creation, duplicate key handling, search functionality, persistence, transaction commits, reset, and getAll behaviors.

PR #698, opened 2025-07-01
Make common and reporter optional and enforce required constraint based on schema

snehagunta:
## Describe your changes

- In v1beta2, we cannot impose required constraint of reporter or common representation, without taking the schema into account.
- This is so we can allow for usecases where reporters can only report reporter or common representations sometimes, but not both. More details are documented in this [DDR](https://docs.google.com/document/d/10FcI3uppt8MJcDKvlHQCQ7h_ujBTw-HvrgEABdxEVbs/edit?tab=t.0#heading=h.v29f3ct0fjjn) under validation.
- Something else to keep in mind is that the current validation rules do not take the state of the existing Resource into account and I will change that in a follow on PR. 
- This PR is only to relax the rules from the API perspective, so I can test the full proposed algorithm (being implemented in [this PR](https://github.com/project-kessel/inventory-api/pull/685)
- Also trying to keep PR size small, so I am making small upfront changes where possible

Fixes #

## Checklist

### DIFF
--- a/internal/middleware/schema_loader.go
+++ b/internal/middleware/schema_loader.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/spf13/viper"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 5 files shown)


tonytheleg (approved):
I have limited background here but since its mostly validation related changes to unblock LGTM

PR #712, opened 2025-07-09
adds monitoring only target for compose plus doc updates

tonytheleg:
## Describe your changes

* Adds new monitoring-only target in docker compose to spin up just the monitoring stack
* Adds update to docker compose doc for the new option
* updates prometheus scrape config to also support scraping metrics for pods running in ephemeral (currently inventory api, KIC, and kessel kafka connect)


tylercreller (approved):
LGTM

PR #790, opened 2025-08-06
Make Representation a defined type just like other tiny types

snehagunta:
## Describe your changes

- ...

Fixes #

## Checklist

### DIFF
--- a/internal/biz/model/common.go
+++ b/internal/biz/model/common.go
@@ -112,6 +112,10 @@ func DeserializeLocalResourceId(value string) LocalResourceId {
 	return Deserialize[LocalResourceId](value)
 }
 
+func DeserializeRepresentation(value JsonObject) Representation {
+	return Representation(value)
+}
+
 // Helper for types that need special increment behavior
 type Incrementable interface {
 	Increment() Incrementable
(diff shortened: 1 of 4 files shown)


internal/biz/model/reporter_representation.go
```diff
@@ -91,10 +94,13 @@ func NewReporterDeleteRepresentation(
 		reporterVersion = &rv
 	}
 
+	representation, err := NewRepresentation(make(JsonObject))
```
akoserwal, line 97: Every delete operation now allocates an empty map instead of using nil, causing memory waste.

internal/biz/model/common.go
```diff
@@ -112,6 +112,10 @@ func DeserializeLocalResourceId(value string) LocalResourceId {
 	return Deserialize[LocalResourceId](value)
 }
 
+func DeserializeRepresentation(value JsonObject) Representation {
+	return Representation(value)
```
akoserwal, line 116: Serialization/Deserialization logic directly into domain models, which breaks the fundamental principle of keeping domain models pure and infrastructure-agnostic.

internal/biz/model/reporter_representation.go
snehagunta: Good point, I will fix this.

snehagunta: @akoserwal so this maybe a byproduct of me breaking up the PRs, but in my final impl, the initialization of this struct would look like this, which does the nil initialization `Representation:     Representation(nil),`

`func NewReporterDeleteRepresentation(
	reporterResourceID ReporterResourceId,
	version Version,
	generation Generation,
	commonVersion Version,
	reporterVersion *ReporterVersion,
) (ReporterDeleteRepresentation, error) {
	if reporterResourceID.UUID() == uuid.Nil {
		return ReporterDeleteRepresentation{}, fmt.Errorf("%w: ReporterResourceId", ErrInvalidUUID)
	}

	return ReporterDeleteRepresentation{
		ReporterRepresentation: ReporterRepresentation{
			Representation:     Representation(nil),
			reporterResourceID: reporterResourceID,
			version:            version,
			generation:         generation,
			commonVersion:      commonVersion,
			reporterVersion:    reporterVersion,
			tombstone:          NewTombstone(true),
		},
	}, nil
}`

So your concern will be addressed in a follow on PR

PR #805, opened 2025-08-12
updates heartbeat script to remove before creating to clean up relation

tonytheleg:
## Describe your changes
Updates the heartbeat fix script to remove the resource first to clean up the relation, then create again


tonytheleg:
Change is not required as the same resource is updated and no extra relations are created

PR #816, opened 2025-08-26
v1beta2: remove strict regex from reporter reference instance_id

Adam0Brien:
## Describe your changes
- while making changes to hbi, I noticed that the instance_id will fail on regex check every time because instance_id should be a uuid
- drops the strict regex check from instance_id 
- effected Check, CheckForUpdate and DeleteResource
```py
def DeleteHost(self, id: str):
        request = delete_resource_request_pb2.DeleteResourceRequest(
            reference=resource_reference_pb2.ResourceReference(
                resource_type="host",
                resource_id=id,
                reporter=reporter_reference_pb2.ReporterReference(
                    type="hbi",
                    instance_id= "3c4e2382-26c1-11f0-8e5c-ce0194e9e144" # previously invalid
                )
            )
        )
```

Fixes #

## Checklist

## Summary by Sourcery

Loosen reporter_reference.instance_id validation by replacing the strict regex with a pattern that accepts both UUIDs and hostnames, regenerate the proto code, and expand tests to cover valid and invalid instance_id formats.

Enhancements:
- Drop the strict regex constraint on reporter_reference.instance_id to allow both UUIDs and hostnames.

Build:
- Regenerate reporter_reference.pb.go after updating the proto definition to include the new regex pattern.

Tests:
- Update default instance_id in existing tests to a valid UUID and add new test cases for valid hostnames and various invalid hostname and UUID formats using protovalidate for validation.


api/kessel/inventory/v1beta2/reporter_reference.proto
```diff
@@ -10,5 +10,5 @@ option java_package = "org.project_kessel.api.inventory.v1beta2";
 
 message ReporterReference {
   string type = 1 [(buf.validate.field).string.min_len = 1, (buf.validate.field).string.pattern = "^[A-Za-z0-9_]+$"];
-  optional string instance_id = 2 [(buf.validate.field).string.pattern = "^[A-Za-z0-9_]+$"];
+  optional string instance_id = 2;
```
akoserwal, line 13: If the instance_id is `redhat`, then removing the complex uuid make sense

snehagunta: If I am understanding the issue correctly, the current validation will not allow UUIDs and in HBI's case, the "." in the instance_id that we recommended to them, "redhat.com". 

What do you think of expanding the validation to allow these values, instead of completely removing them? This will protect us against empty strings, spaces and other special characters like : ,#, @ which if present might cause issues when we start sending this value to spicedb.

It would be a breaking change later if we want to add these rules later to the API contract, so my recommendation is to start strict and relax rules as needed.

api/kessel/inventory/v1beta2/reporter_reference.pb_test.go
```diff
@@ -8,49 +8,49 @@ import (
 )
 
 func TestReporterReference_BasicFields(t *testing.T) {
-	instID := "hbi"
+	instID := "3c4e2382-26c1-11f0-8e5c-ce0194e9e144"
```
snehagunta, line 11: Can you also add tests with the "redhat.com" scenario? To test that we are allowing dots.

api/kessel/inventory/v1beta2/reporter_reference.proto
Adam0Brien: Makes perfect sense! I added two regex checks to the proto, one for the UUID case (invalid UUID's will also fail), and another to allow for strings like "redhat" or "redhat.com"

I've also added a few more tests cases for UUID's and Strings

PR #834, opened 2025-09-04
Capture validation errors for metrics

tylercreller:
- Moves validation middleware after metrics bootstrapping to make sure we are capturing validation error codes


tonytheleg (approved):
LGTM

PR #844, opened 2025-09-08
[WIP] Rhcloud 41014 - Fix replication to delete + create tuples when the id -> workspaceId changes

Adam0Brien:
Fixes #https://issues.redhat.com/browse/RHCLOUD-41014

## Checklist

@sourcery-ai summary


internal/biz/usecase/resources/resource_service.go
```diff
@@ -822,3 +841,91 @@ func computeReadAfterWrite(uc *Usecase, write_visibility v1beta2.WriteVisibility
 	}
 	return !common.IsNil(uc.ListenManager) && uc.Config.ReadAfterWriteEnabled && isSPInAllowlist(reporterPrincipal, uc.Config.ReadAfterWriteAllowlist)
 }
+
+// getCurrentWorkspaceIdFromDatabase queries the database for the current workspace ID of a resource
+func (uc *Usecase) getCurrentWorkspaceIdFromDatabase(tx *gorm.DB, reporterResourceKey model.ReporterResourceKey) (string, error) {
```
snehagunta, line 846: 1. Any interaction with the DB should always be in the repository layer. The resource_service is an application service, so it should only have cross cutting concerns like interaction between domain models, transactions security etc. 
2. We already have a find repository method to lookup an existing resource, so I would be very careful about adding specific find methods unless absolutely necessary.
3. workspaceId to localResourceId is one of the many relations that we will need to replicate to spicedb, so this implementation is very specific to the current requirement, but our implementation needs to be such that we can calculate multiple tuples for a given resource.

internal/biz/model_legacy/eventoperationtype.go
```diff
@@ -15,3 +20,26 @@ const (
 func (o eventOperationType) OperationType() eventOperationType {
 	return o
 }
+
+// Value implements the driver.Valuer interface for database storage
+func (o eventOperationType) Value() (driver.Value, error) {
+	return string(o), nil
+}
+
+// Scan implements the sql.Scanner interface for database retrieval
```
snehagunta, line 29: This is a smell to me, why do you need to add a DB scan to an enum?

internal/data/outboxeventsrepository.go
```diff
@@ -18,3 +20,38 @@ func PublishOutboxEvent(tx *gorm.DB, event *model_legacy.OutboxEvent) error {
 	}
 	return nil
 }
+
+// HandleWorkspaceRelationOutboxEvents generates both create and delete outbox events for workspace relation changes
+// This ensures new workspace relations are added before old ones are deleted, preventing access gaps
+// Following the add-first-then-delete approach: brief period of extra access is acceptable
+func HandleWorkspaceRelationOutboxEvents(tx *gorm.DB, resource model_legacy.Resource, oldWorkspaceId string, namespace string, txid string) error {
```
snehagunta, line 27: Repository layer should not have any business logic, it should just handle simple reads and writes to the database. This code should not live here.

snehagunta:
@Adam0Brien I am thinking it maybe easier to start fresh with this branch, since the patterns are inconsistent with what we want to follow and it maybe harder to fix. Let's meet tomorrow to figure a smaller starting place so we can evolve the solution bit by bit.

PR #890, opened 2025-09-22
adds mutex locks to avoid offset commit race conditions + test updates

tonytheleg:
## Describe your changes
* Implements bug fix found in Kessel Inventory Consumer where there is a race condition when commiting offsets during shutdown/recovation (See https://github.com/project-kessel/inventory-consumer/pull/148 for more details)
* Updates tests for changes:
  * Adds relevant tests for the race condition
  * Updates `New()` consumer function call to accept a consumer as an argument, allowing for better mocking
  * Updates tests to pass mock consumer to `New()` 
  * Updates mock consumer function calls to only be expected when offsets are present

### DIFF
--- a/internal/consumer/consumer.go
+++ b/internal/consumer/consumer.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"slices"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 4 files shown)


internal/consumer/consumer.go
```diff
@@ -620,6 +671,26 @@ func (i *InventoryConsumer) RebalanceCallback(consumer *kafka.Consumer, event ka
...
+		// Check if shutdown is already in progress to avoid double commits
+		i.offsetMutex.Lock()
+		shutdownInProgress := i.shutdownInProgress
+		hasOffsets := len(i.OffsetStorage) > 0
+		i.offsetMutex.Unlock()
+
+		if shutdownInProgress {
+			i.Logger.Info("shutdown in progress, skipping rebalance offset commit")
+			i.lockToken = ""
+			i.lockId = ""
```
tylercreller, line 683: Should we move the resetting of these vars to the shutdown func? It seems "smelly" to reset `lockToken` and `lockId` as a side effect of a rebalance if a shutdown is in progress. It's not immediately apparent to me if we _need_ to reset these here.

lennysgarage (approved):
lgtm 👍

PR #902, opened 2025-09-25
Generation support and make deletes idempotent

snehagunta:
Make tombstoned resources found; multiple deletes are just version updates; generation support

## Describe your changes

- ...

Fixes #

## Checklist

## Summary by Sourcery

Add support for resource generation tracking and make delete operations idempotent while preserving historical (tombstoned) records. Remove automatic tombstone filtering in repository lookups, enforce composite unique constraints in the fake repository, and implement case-insensitive key matching. Expand the test suite with comprehensive contract and unit tests covering create/update/delete lifecycles, idempotency, unique constraint enforcement, and generation increment logic.

New Features:
- Support generation increment and representation version reset when reviving tombstoned resources
- Treat delete operations as idempotent by updating representation version rather than failing on duplicate deletes
- Allow FindResourceByKeys to return tombstoned resources instead of filtering them out

Enhancements:
- Remove default tombstone filter from GORM queries to include deleted records
- Enforce composite unique constraints and proper version/generation ordering in the fake repository
- Implement case-insensitive matching for resource lookup keys

Tests:
- Add extensive contract tests for real and fake repositories covering lifecycle workflows, idempotency, unique constraints, and key casing
- Add unit tests for Generation.Increment behavior

### DIFF
--- a/internal/data/resource_repository.go
+++ b/internal/data/resource_repository.go
@@ -211,7 +211,6 @@ func (r *resourceRepository) FindResourceByKeys(tx *gorm.DB, key bizmodel.Report
 	query = query.Where("LOWER(rr.local_resource_id) = LOWER(?)", key.LocalResourceId().Serialize())
 	query = query.Where("LOWER(rr.resource_type) = LOWER(?)", key.ResourceType().Serialize())
 	query = query.Where("LOWER(rr.reporter_type) = LOWER(?)", key.ReporterType().Serialize())
-	query = query.Where("rr.tombstone = ?", false)
 
 	// Only add reporter_instance_id condition if it's not empty
 	if reporterInstanceId := key.ReporterInstanceId().Serialize(); reporterInstanceId != "" {
--- a/internal/biz/model/common.go
+++ b/internal/biz/model/common.go
@@ -279,6 +279,13 @@ func (g Generation) Uint() uint {
 	return uint(g)
 }
 
+// Increment returns a new Generation with the value incremented by 1.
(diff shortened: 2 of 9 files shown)


internal/data/resource_repository.go
```diff
@@ -211,7 +211,6 @@ func (r *resourceRepository) FindResourceByKeys(tx *gorm.DB, key bizmodel.Report
 	query = query.Where("LOWER(rr.local_resource_id) = LOWER(?)", key.LocalResourceId().Serialize())
 	query = query.Where("LOWER(rr.resource_type) = LOWER(?)", key.ResourceType().Serialize())
 	query = query.Where("LOWER(rr.reporter_type) = LOWER(?)", key.ReporterType().Serialize())
-	query = query.Where("rr.tombstone = ?", false)
```
tylercreller, line 214: I think removing this will cause consecutive delete calls to the same resource to return a 200 and increment `representation_version` arbitrarily even when it is already tombstoned.

1. Report Resource (version 0, tombstone=f)
2. Delete Resource (version 1, tombstone=t)
3. Delete Resource (version 2, tombstone=t)

snehagunta: Yeah, but the issue is if I do not exclude soft deleted resource, it will create a new resource and a follow on delete will fail the unique constraint because it will look exactly the same as the previous delete. That was the issue to start with.

snehagunta: I do have to test with check and check for update with this removal, adding tests for that now.

PR #1339, opened 2026-05-05
RHCLOUD-47148: minor bug fix when providing custom schema for full kessel compose

tonytheleg:
## Describe your changes
Found a bug in the full kessel setup when trying to provide a custom schema file during testing for RHCLOUD-47148
The .env file explicitly sets `SCHEMA_ZED_FILE=""`. The script then sources this file and overwrites the environment variable passed in, clobbering it

This fixes the issue by only letting the .env file set defaults, not override explicit environment variables.

For RHCLOUD-47148

## Summary by CodeRabbit

* **Bug Fixes**
  * Startup script now treats .env entries as default values and preserves any environment variables already set before launch. This prevents overwriting user-provided configuration during initialization, ensuring custom settings supplied in the shell are respected and retained for more reliable and predictable startups.

### DIFF
--- a/scripts/start-full-kessel.sh
+++ b/scripts/start-full-kessel.sh
@@ -6,11 +6,16 @@ source ./scripts/check_docker_podman.sh
 COMPOSE_DIR="development/full-kessel"
 ENV_FILE="${COMPOSE_DIR}/.env"
 
-# Load .env for schema handling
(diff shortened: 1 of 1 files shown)


lennysgarage (approved):
LGTM

PR #1342, opened 2026-05-08
RHCLOUD-47590: updates the test ephem deployment for latest changes including new auth config

tonytheleg:
## Describe your changes
* updates `development/kessel-inventory-ephem-w-auth-test.yaml` for changes in authn configuration, as well as numerous other updates that are in the standard ephemeral deployment (schema changes, WAL mode, etc)

This is part of updating the [Using Authentication](https://project-kessel.github.io/docs-internal-overlay/for-red-hatters/using-kessel/inventory-api/using-authentication/) guide in the internal docs. Looking to leverage this template to simplify the process for SP's wanting to test their auth credentials in ephemeral before rolling out to stage

For RHCLOUD-47590

## Summary by CodeRabbit

* **Chores**
  * Switched API authentication to a chained authenticator with OIDC plus identity fallback.
  * Changed schema storage to in-memory backed resources and removed a consistency flag.
  * Switched public service to an h2c-enabled endpoint and adjusted application versioning/dependencies.
  * Removed Kafka Connect / Debezium outbox connectors and cleaned up related template parameters; added a new API config parameter.

### DIFF
--- a/development/kessel-inventory-ephem-w-auth-test.yaml
+++ b/development/kessel-inventory-ephem-w-auth-test.yaml
@@ -170,33 +170,9 @@ parameters:
   - description: Number of replicas
     name: REPLICAS
     value: "1"
-  - name: TOPIC_HEARTBEAT_PREFIX
-    value: debezium-heartbeat
-    description: Prefix for the connector heartbeat topic
-  - name: DEBEZIUM_ACTION_QUERY
-    value: "SELECT pg_logical_emit_message(false, 'heartbeat', now()::varchar);"
-    description: Query action that runs for each heartbeat event
-  - name: DEBEZIUM_HEARTBEAT_INTERVAL_MS
-    value: "300000"
-    description: The interval for the Debezium heartbeat in ms
-  - name: DEBEZIUM_POLL_INTERVAL_MS
-    value: "250"
-    description: The interval for the Debezium batch processing
   - description: Name of the Inventory API Config K8s ConfigMap
     name: INVENTORY_API_CONFIG_NAME
     value: inventory-api-config
-  - description: Memory request for job
-    name: JOB_REQUESTS_MEMORY
-    value: '256Mi'
-  - description: CPU request for job
-    name: JOB_REQUESTS_CPU
-    value: '100m'
-  - description: Memory limit for job
-    name: JOB_LIMITS_MEMORY
-    value: '384Mi'
-  - description: CPU limit for job
-    name: JOB_LIMITS_CPU
-    value: '200m'
   - name: SSO_SERVER_URL
     description: URL for SSO Endpoint to validate clients
     required: true
--- a/development/kessel-inventory-ephem-w-auth-test.yaml
+++ b/development/kessel-inventory-ephem-w-auth-test.yaml
@@ -9,18 +9,34 @@ objects:
       name: inventory-api-config
     data:
       inventory-api-config.yaml: |
-        server:
-          grpc:
-            certfile: /serving-certs/tls.crt
-            keyfile: /serving-certs/tls.key
         storage:
           max-serialization-retries: 10
+          outbox-mode: wal
         authn:
-          oidc:
-            authn-server-url: ${SSO_SERVER_URL}
-            client-id: ${INVENTORY_CLIENT_ID}
-            skip-client-id-check: true
-            insecure-client: false
+          authenticator:
+            type: first_match
+            chain:
+              - type: oidc
+                enable: true
+                transport:
+                  http: false
+                  grpc: true
+                config:
+                  insecure-client: true
+                  authn-server-url: ${SSO_SERVER_URL}/auth/realms/redhat-external
+                  skip-client-id-check: true
+                  skip-issuer-check: true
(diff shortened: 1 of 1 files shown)


development/kessel-inventory-ephem-w-auth-test.yaml
```diff
@@ -226,26 +182,21 @@ parameters:
   - name: DEBEZIUM_POLL_INTERVAL_MS
     value: "250"
     description: The interval for the Debezium batch processing
```
lennysgarage: We can probably also remove these env variables now that the connector is removed

development/kessel-inventory-ephem-w-auth-test.yaml
```diff
@@ -9,18 +9,34 @@ objects:
...
+            type: first_match
+            chain:
+              - type: oidc
+                enable: true
+                transport:
+                  http: false
+                  grpc: true
+                config:
+                  insecure-client: true
+                  authn-server-url: ${SSO_SERVER_URL}/auth/realms/redhat-external
```
lennysgarage, line 26: `${SSO_SERVER_URL}` is just the base SSO url now then?

development/kessel-inventory-ephem-w-auth-test.yaml
```diff
@@ -117,7 +75,7 @@ objects:
       envName: ${ENV_NAME}
       database:
         name: kessel-inventory
-        version: 17
+        version: 16
```
lennysgarage, line 78: Is postgres 17 still not supported in ephm environments?

development/kessel-inventory-ephem-w-auth-test.yaml
```diff
@@ -179,27 +138,24 @@ objects:
...
             public:
               enabled: true
-              apiPath: inventory
-              tls: true
+              h2cEnabled: true
+              h2cTargetPort: 9000
+              apiPath: kessel
       testing:
         iqePlugin: kessel-inventory
+
```
lennysgarage, line 158: If we're updating this ephm template up with the others should we also add the `business-metrics-collect` cronjob? I see the non auth ephm deploy has it

lennysgarage: I think we should considering we've added the environment variables for it `JOB_LIMITS_CPU` and such. https://github.com/project-kessel/inventory-api/blob/main/deploy/kessel-inventory-ephem.yaml#L217-L219

lennysgarage (commented):
Left a few comments for syncing things up with some of the other deploy files!

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: yep, missed those, thank you!

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: correct, this is pulled from our inventory config secret in app interface under stage and prod. 
```
            config:
              insecure-client: true
              authn-server-url: {{{ auth_url }}}/realms/redhat-external
```

I left the use of `SSO_SERVER_URL` instead of `auth_url` since it was already in the template and the updated authentication guide when its done will provide the exact URL to use for stage

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: Correct, 16 is the latest available:
https://github.com/RedHatInsights/clowder/blob/master/config/crd/bases/cloud.redhat.com_clowdapps.yaml#L119-L126

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: to be fair, we dont actually need the job settings either, thats a copy/paste i didnt consider. I took the metrics off because the sole purposes of this deployment is focused on testing auth, the same could be done for the job settings as well. the only reason business metrics and job setting stuff is in the main ephemeral deployment is because we may need to test those specific functions, i wouldn't see us using this alternate version to test anything but auth. What do you think, if you think there is still value in adding them, i can do so, its easy enough

development/kessel-inventory-ephem-w-auth-test.yaml
lennysgarage: Makes sense!

development/kessel-inventory-ephem-w-auth-test.yaml
lennysgarage: > the same could be done for the job settings as well

I think removing them to keep this deploy focused on auth makes sense

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: Updated

development/kessel-inventory-ephem-w-auth-test.yaml
tonytheleg: Updated

lennysgarage (approved):
LGTM!

PR #1381, opened 2026-06-22
RHCLOUD-48646: Add support for optional ephemeral x-rh-identity heade…

tylercreller:
- Enable checkself in ephemeral by adding x-rh-identity to the authn chain and configuring selfsubjectstrategy
- Backwards compatible, services without the header still fall through to allow-unauthenticated

Test results
```
  ┌───────────┬──────────┬───────────────┬───────────────────────────────────────────────┐
  │   Test    │ Protocol │   Identity    │                    Result                     │
  ├───────────┼──────────┼───────────────┼───────────────────────────────────────────────┤
  │ checkself │ HTTP     │ none          │ 403 meta authorization denied (correct)       │
  │           │ :8000    │               │                                               │
  ├───────────┼──────────┼───────────────┼───────────────────────────────────────────────┤
  │ checkself │ HTTP     │ x-rh-identity │ ALLOWED_FALSE + consistency token     │
  │           │ :8000    │               │                                               │
  ├───────────┼──────────┼───────────────┼───────────────────────────────────────────────┤
  │ check     │ gRPC     │ none          │ ALLOWED_FALSE + consistency token (backwards  │
  │           │ :9000    │               │ compatible)                                   │
  ├───────────┼──────────┼───────────────┼───────────────────────────────────────────────┤
  │           │ HTTP     │               │ 403 meta authorization denied (correct —      │
  │ check     │ :8000    │ none          │ check is a service-to-service op, should use  │
  │           │          │               │ gRPC)                                         │
  └───────────┴──────────┴───────────────┴───────────────────────────────────────────────┘
  
                            
```

## Summary

* **Chores**
  * Updated the service’s authentication handling to use identity headers first, with an automatic fallback to unauthenticated access.
  * Enhanced the authorization setup by enabling a Red Hat RBAC self-subject strategy, including the required OIDC issuer domain configuration for validation.

### DIFF
--- a/deploy/kessel-inventory-ephem.yaml
+++ b/deploy/kessel-inventory-ephem.yaml
@@ -13,11 +13,26 @@ objects:
           max-serialization-retries: 10
           outbox-mode: wal
         authn:
-          allow-unauthenticated: true
+          authenticator:
+            type: first_match
+            chain:
+              - type: x-rh-identity
+                transport:
+                  http: true
+                  grpc: false
+              - type: allow-unauthenticated
         authz:
           kessel:
             insecure-client: true
             enable-oidc-auth: false
+        selfsubjectstrategy:
+          redhatRbacSelfSubjectStrategy:
+            enabled: true
+            issuerDomain: redhat
+            # oidcIssuerDomains is unused in ephemeral (no OIDC in the authn chain) but required by validation
+            oidcIssuerDomains:
+              - iss: https://not-used.example.com
+                domain: redhat
         consumer:
           enabled: true
           topic: outbox.event.kessel.tuples


deploy/kessel-inventory-ephem.yaml
```diff
@@ -13,11 +13,25 @@ objects:
...
+              - type: allow-unauthenticated
         authz:
           kessel:
             insecure-client: true
             enable-oidc-auth: false
+        selfsubjectstrategy:
+          redhatRbacSelfSubjectStrategy:
+            enabled: true
+            issuerDomain: redhat
+            oidcIssuerDomains:
```
Rajagopalan-Ranganathan, line 33: i am not sure if the issuer domain will be sso stage - its a local keycloak instance AFAIK in the same NS.

tylercreller: Yeah good catch, I actually believe this config is not doing anything useful here and we just need the `enabled` and `issuerDomain`. Validating...

tylercreller: It looks like the configuration is required but is not actually being used in ephemeral. I've set it to an obvious foobar value for now `https://not-used.example.com`

Rajagopalan-Ranganathan: yeah exactly was going to suggest that, would be good to add a TODO there saying it should be keycloak issuer domain in the NS of the deployment if any.

Rajagopalan-Ranganathan (approved):

PR #1383, opened 2026-06-23
RHCLOUD-48142 Serialized schema

josejulio:
## Describe your changes

- ...

Fixes #

## Checklist


data/schema/resources/k8s_policy.yaml
```diff
@@ -0,0 +1,39 @@
...
+reporters:
+- name: acm
+  description: Advanced Cluster Management policy reporter
+  schema:
+    type: object
+    properties:
+      disabled:
+        type: boolean
+        description: Defines if the policy is currently enabled or disabled across
+          all targets
```
lennysgarage, line 26: Probably more stylistic than anything but this description reads across two lines. Don't know if that has any issues with parsing

josejulio: We can consolidate in a single line, but multiline is supported this way in yaml.

Check the `plain` section and the equivalent to a single string in [here](https://yaml-multiline.info/).

data/schema/resources/sample_app.yaml
```diff
@@ -0,0 +1,185 @@
+# Sample Application Resource - Comprehensive Test Schema
```
lennysgarage: Should this live in the same directory as the others? just that if someone tried to report a resource, it would most likely error as the tuples in spicedb don't exist. I like it as an example tho!

lennysgarage (commented):
Converted schemas look accurate to me! Still reading through how the unified schema reads them but like the format of them so far! Easy to follow

data/schema/resources/sample_app.yaml
josejulio: Probably not, at this point I'm just testing different things- maybe I'll add this as parts of docs, or maybe not! in the end this should be transpiled from the unified schema and thus might be opaque but documented to be understood by humans/agents.

data/schema/resources/role_binding.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - name: role
+      target: rbac/role
+      field: role_id
+      cardinality: one
+
+    # Principal receiving the role
+    - name: principal
+      target: rbac/principal
+      field: principal_id
+      cardinality: one
```
lennysgarage, line 34: I'm wondering for relations that accept a union type like `relation subject: [Any principal or group.member]` how would you model it in the YAML format here?

josejulio: interesting :thinking: 
I suppose we must be told which one it is. is that right? 

We will have to take that as input from the reporter when there are multiple types at play.
I haven't really though about something so this is just a rough idea


```
target:
  - rbac/principal
  - rbac/group#member
target_field: principal_relation # or something else
```
and the `principal_relation` might be defined as
```
principal_relation:
  type: string
  enum: 
    - "rbac/principal"
    - "rbac/group#member"
```

or even something more convenient for the reporter like:

```
target:
  principal: rbac/principal
  team: rbac/group#member
target_field: principal_relation # or something else
```

```
principal_relation:
  type: string
  enum: 
    - "principal"
    - "team"
```

josejulio: Do we have relations accepting unions for any of the work in https://github.com/project-kessel/inventory-api/pull/1382 ? or this is more of a future-thing?
I saw we have relations like this in rbac, but not hbi - unsure about the others.

lennysgarage: No unions for any of the work in #1382 or features that I can tell. More so for the future, we do have the ability to represent in the starlark DSL with `anyOf()` for it, so I wanted to bring it up. But it shouldn't be blocking for features if we don't have it

lennysgarage: > I suppose we must be told which one it is. is that right?

Yeah and i think going down the route of something similar to the mapped target feels right

PR #1396, opened 2026-07-02
moves relations to dependency vs optional

tonytheleg:
## Describe your changes
While Relations is listed as an optional dep, teams that reference kessel-inventory as a dep do not get all optional deps defined by Kessel unless they deploy kessel specifically. This is causing clowder injection to fail setting the kessel URL needed to talk to relations. This addresses the issue

Also moves KKC into deps for the same reaons

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the deployment template so a previously optional backend connection is now required, ensuring the app starts with the expected supporting service available.


tylercreller (approved):
LGTM

PR #1386, opened 2026-06-24
RHCLOUD-48646: Add compose options for Inventory w/ SpiceDB

tylercreller:
- Adds compose file for SpiceDB dependencies
- Adds script for launching local spicedb compose and downloading stage zed schema
  - Adds gitignore for development zed schema
- Adds `inventory-up-w-spicedb` make target
- Adds e2e runbook file
- Bolsters config error handling for relations/spicedb

## Summary by CodeRabbit

* **New Features**
  * Added a one-command `make inventory-up-spicedb` flow to run Inventory with an embedded SpiceDB stack.
  * Added a local Docker Compose configuration for SpiceDB + Postgres + Kafka with schema wiring.
* **Bug Fixes**
  * Improved configuration validation to fail fast when the selected authorization backend is missing.
  * Enabled relations processing for the SpiceDB-backed relations repository.
* **Documentation**
  * Updated the local Docker Compose guide with embedded SpiceDB instructions and ports.
  * Added an embedded SpiceDB end-to-end developer runbook.
* **Chores**
  * Updated local schema/source URLs to use the correct `project-kessel/rbac-config` locations.
  * Updated ignore rules for the local embedded schema file.

PR #1405, opened 2026-07-08
RHCLOUD-48927: Add SpiceDB direct mode parameters to deployment template

tylercreller:
## What changed

- Added two optional volume mounts to the deployment template: the spicedb-schema configmap at /etc/spicedb-schema and the spicedb-config secret at /etc/spicedb-secret (readOnly). Both use optional: true so pods start normally in environments where these resources don't exist.
- Added two template parameters (SPICEDB_SECRET_NAME, SPICEDB_SCHEMA_CONFIGMAP) to keep the mounted resource names configurable.

## Why

This is the deployment side of connecting inventory-api directly to SpiceDB as part of merging the relations service. The Go code for SpiceDB direct mode was wired in via #1386 and #1380 but the non-ephemeral template had no way to provide the schema file or preshared key to the process.

All SpiceDB configuration (impl, endpoint, token-file, schema path, TLS, consistency) is controlled through the inventory-api-config secret rather than template env vars. Viper gives env vars higher precedence than config file values, so using env vars would prevent the config secret from being the single control point. This matches how the existing kessel authz settings (sa-client-id, sso-token-endpoint, etc.) are already managed.

## Validation

- Processed the template locally with oc process using default parameters. Output shows the two optional volume mounts alongside existing resources, with the three original env vars unchanged.
- Verified optional: true on both the secret and configmap volumes prevents pod startup failures when the resources are absent.

Jira: RHCLOUD-48927

Depends on #1386 and #1380 being merged for the SpiceDB code path to be reachable at runtime. The template change itself is safe to merge independently. A follow-up app-interface MR will set authz.impl to spicedb in the stage config secret and add the spicedb subsection (endpoint, token-file, schema-file, use-tls, fully-consistent) to activate direct mode.

## Summary by CodeRabbit

* **New Features**
  * Added support for optionally mounting a SpiceDB schema ConfigMap into the application container.
  * Added support for optionally mounting a SpiceDB secret into the application container as read-only.

* **Chores**
  * Extended deployment template parameters to customize the schema ConfigMap and secret names used for these mounts.

### DIFF
--- a/deploy/kessel-inventory.yaml
+++ b/deploy/kessel-inventory.yaml
@@ -85,6 +85,11 @@ objects:
                   mountPath: "/data/schema/resources"
                 - name: resources-tarball
                   mountPath: "/mnt/resources"
+                - name: spicedb-schema
+                  mountPath: "/etc/spicedb-schema"
+                - name: spicedb-secret
+                  mountPath: "/etc/spicedb-secret"
+                  readOnly: true
             volumes:
               - name: config-volume
                 secret:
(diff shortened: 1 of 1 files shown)


lennysgarage (approved):
LGTM

PR #1413, opened 2026-07-14
Add PR performance regression test workflow

larrymou9:
## What changed
Added `.github/workflows/perf-test.yml` — a GitHub Actions workflow that triggers an automated performance regression test on every PR.

## Why
Enables early detection of performance regressions by running a Jenkins perf test pipeline against the Konflux-built PR image before merging.

## Validation
- Workflow uses `pull_request_target` so fork PRs can access repo secrets
- All GitHub context values passed through `env:` to prevent shell injection
- `jq --arg` used to safely build the JSON payload
- `GITHUB_TOKEN` permissions set to `{}` (least privilege — no GitHub API access needed)
- `GOSMEE_WEBHOOK_SECRET` already set as a repo secret ✅

## Jira
https://redhat.atlassian.net/browse/HCEPERF-780

## Summary by CodeRabbit

* **New Features**
  * Added automated performance regression testing for pull requests.
  * Performance tests now wait for the corresponding build image, trigger the test run, and report a monitoring link.
  * Duplicate runs are automatically canceled when a newer update is available.

### DIFF
--- a/.github/workflows/perf-test.yml
+++ b/.github/workflows/perf-test.yml
@@ -1,7 +1,4 @@
 ---
-# Copy this file to project-kessel/inventory-api as:
-#   .github/workflows/perf-test.yml
-#
 # Required GitHub secret (set at org or repo level):
 #   GOSMEE_WEBHOOK_SECRET  — shared HMAC secret, must match the value in the
 #                            Gosmee client configmap on the perf cluster.
--- a/.github/workflows/perf-test.yml
+++ b/.github/workflows/perf-test.yml
@@ -21,6 +21,7 @@ jobs:
   trigger-perf-test:
     name: Trigger Jenkins Performance Test
     runs-on: ubuntu-latest
+    permissions: {}  # this job only calls external services; no GitHub API access needed
     steps:
       - name: Resolve image to test
         id: image
--- a/.github/workflows/perf-test.yml
+++ b/.github/workflows/perf-test.yml
@@ -12,6 +12,10 @@ name: Performance Regression Test
 on:
   pull_request:
     types: [opened, synchronize, reopened]
+    paths-ignore:
+      - '.github/**'
+      - '**.md'
+      - 'docs/**'
 
 jobs:
   trigger-perf-test:
--- a/.github/workflows/perf-test.yml
+++ b/.github/workflows/perf-test.yml
@@ -100,15 +100,14 @@ jobs:
           echo "SHA:       ${PR_SHA}"
           echo "Image tag: on-pr-${PR_SHA}"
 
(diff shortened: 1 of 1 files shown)


tylercreller:
@coderabbitai full review

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,118 @@
...
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths-ignore:
+      - '.github/**'
+      - '**.md'
+      - 'docs/**'
+
+jobs:
+  trigger-perf-test:
```
tylercreller, line 22: Quick pushes to a PR will stack up runs polling Quay for 30 minutes each, firing off Jenkins jobs for commits nobody cares about anymore. Worth adding a concurrency group on the PR number with cancel-in-progress so only the latest commit gets tested.

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,118 @@
+---
+# Copy this file to project-kessel/inventory-api as:
+#   .github/workflows/perf-test.yml
```
tylercreller: Lines 2-3 say "Copy this file to project-kessel/inventory-api" but it's already here. Leftover from development, should be removed.

tylercreller (changes requested):

.github/workflows/perf-test.yml
```diff
@@ -0,0 +1,118 @@
...
+#
+# Required GitHub secret (set at org or repo level):
+#   GOSMEE_WEBHOOK_SECRET  — shared HMAC secret, must match the value in the
+#                            Gosmee client configmap on the perf cluster.
+#
+# See ci-configs/jenkins/gh-pr-testing/README.md for the full onboarding guide.
```
tylercreller: This path does not exist in this repo, we should remove this or link the full repo path

tylercreller:
@larrymou9 Can you update your description to match our default pattern? https://github.com/project-kessel/inventory-api/blob/main/.github/pull_request_template.md

.github/workflows/perf-test.yml
larrymou9: Thanks Tyler,  I have made the change

.github/workflows/perf-test.yml
larrymou9: Thank you! I've removed it

.github/workflows/perf-test.yml
larrymou9: This is a very valid concern and thank you very much for pointing it out. I have made the respective change :)

tylercreller (approved):
LGTM

